PR #6, opened 2019-01-23
use updated jupyter session proxy and update rstudio version

### LINKED ISSUE #4: Update R & Rstudio versions
## RStudio server

Is it possible to update

https://github.com/SwissDataScienceCenter/renku-jupyter/blob/master/docker/r/Dockerfile#L24

to the latest
```bash
ENV RSTUDIO_PKG=rstudio-server-1.1.463-amd64.deb
```
?

(Plots don't display properly in the RStudio panel otherwise).

## R via conda vs apt

R is installed via conda: https://github.com/SwissDataScienceCenter/renku-jupyter/blob/master/docker/r/Dockerfile#L37.

Is it necessary?  Would and apt install work?  (I'm trying to figure out how dependencies will/should be handled in the R world.)

## R version

Can we update to the latest R (via Conda or apt)?  Does r-base point the latest by default if we don't specify a version number?
```bash
conda install -c r r-base
```

### LINKED ISSUE #5: fix the fallout from nbserverproxy project rename
The project is now https://github.com/jupyterhub/jupyter-server-proxy. The R image Dockerfile needs to be fixed.

rokroskar:
closes #4 
closes #5

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -21,7 +21,7 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 # You can use rsession from rstudio's desktop package as well.
-ENV RSTUDIO_PKG=rstudio-server-1.0.136-amd64.deb
+ENV RSTUDIO_PKG=rstudio-server-1.1.463-amd64.deb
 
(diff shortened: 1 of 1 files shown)


rokroskar:
because it's easy 👍 

what do you suggest?

PR #29, opened 2019-04-10
pre-stop and post-start pod lifecycle scripts

jachro:
This is a jupyter part of the issue raised for `renku-notebooks` (https://github.com/SwissDataScienceCenter/renku-notebooks/issues/126)
It introduces `pre-stop` and `post-start` scripts to be executed by Kubernetes when jupyter container got created and before killing.

### DIFF
--- a/docker/base/lifecycle/pre-stop.sh
+++ b/docker/base/lifecycle/pre-stop.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+GIT_STATUS=`git status -s`
+if [ -z "$GIT_STATUS" ] ; then
+  exit 0
+fi
+
+CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
+CURRENT_SHA=`git rev-parse --short HEAD`
+AUTOSAVE_BRANCH="renku/autosave/$JUPYTERHUB_USER/${CURRENT_BRANCH}/${CURRENT_SHA}"
+git checkout -b "$AUTOSAVE_BRANCH"
+git add .
+git commit -am "Auto-saving for $JUPYTERHUB_USER on branch $CURRENT_BRANCH and commit $CURRENT_SHA"
+git push origin "$AUTOSAVE_BRANCH"
+git checkout master
+git branch -D "$AUTOSAVE_BRANCH"
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -64,6 +64,8 @@ RUN pipx install ${RENKU_PIP_SPEC} --pip-args="--pre" renku && \
 COPY git-config.bashrc /home/$NB_USER/
 RUN cat /home/$NB_USER/git-config.bashrc >> /home/$NB_USER/.bashrc && rm /home/$NB_USER/git-config.bashrc
 
+COPY lifecycle/ /usr/local/bin/
+
 # Add user `jovyan` to sudoers (passwordless)
 USER root
 RUN echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \


docker/base/lifecycle/pre-stop.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+GIT_STATUS=`git status -s`
+if [ -z "$GIT_STATUS" ] ; then
+  exit 0
+fi
+
+CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
+AUTOSAVE_BRANCH="auto-save-$JUPYTERHUB_USER-$CURRENT_BRANCH"
```
rokroskar: can we add the commit-sha here? Then you would also know more specifically what the base was

docker/base/lifecycle/post-start.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -z "$AUTOSAVE_REMOTE_BRANCH" ] ; then
+  echo No auto-save branch
+  exit 0
+fi
+
+IFS='/' AUTOSAVE_REMOTE_BRANCH_ITEMS=( $AUTOSAVE_REMOTE_BRANCH )
+
+if [ "${#AUTOSAVE_REMOTE_BRANCH_ITEMS[@]}" -lt 3 ] ; then
+  echo "Auto-save branch in wrong format; cannot recover to the state from that branch"
```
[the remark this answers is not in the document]
jachro: Changed

docker/base/lifecycle/pre-stop.sh
jachro: SHA added

docker/base/lifecycle/post-start.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+AUTOSAVE_BRANCH_PREFIX="auto-save-$JUPYTERHUB_USER"
```
jirikuncar: It would be nice to group the branches with renku prefix since there might isolation branches too:
```suggestion
AUTOSAVE_BRANCH_PREFIX="renku/autosave/${SERVER_NAME}"
```

docker/base/lifecycle/pre-stop.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+GIT_STATUS=`git status -s`
+if [ -z "$GIT_STATUS" ] ; then
+  exit 0
+fi
+
+CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
+CURRENT_SHA=`git rev-parse HEAD`
+AUTOSAVE_BRANCH="auto-save-$JUPYTERHUB_USER-$CURRENT_BRANCH-$CURRENT_SHA"
```
jirikuncar:
```suggestion
AUTOSAVE_BRANCH_PREFIX="renku/autosave/${SERVER_NAME}/${CURRENT_BRANCH}/${CURRENT_SHA}"
```

jachro: Right, I quite like that. It brings some nice order.

rokroskar: @jirikuncar does git treat `/` in a special way when it stores the refs? Or is that just cosmetic.

jachro: I was asking myself the same question and I got a thought that I simply test it. The only problem I see is tokenizing the branch name could be a bit more difficult.

docker/base/lifecycle/post-start.sh
rokroskar: why `SERVER_NAME`? it's a function of `namespace`, `project` and `commit-sha` -- we need to be able to identify the user

jirikuncar: namespace and project are already part of the remote where you are pushing the branch

rokroskar: I know -- what I'm asking is why you want to include `SERVER_NAME`

jirikuncar: if you relaunch the server from the same sha1 will the server name be the same?

jirikuncar: if you relaunch the server from the same sha1 will the server name be the same?

rokroskar: yes - but I would prefer to keep the sha and original branch name as part of the autosave branch ref to be easier to identify

jirikuncar: fine, just use / as separator for nice branch grouping to dirs

docker/base/lifecycle/pre-stop.sh
jirikuncar: yes, the refs are stored in directories - hence good option when there are many autosave branches

docker/base/lifecycle/post-start.sh
rokroskar: 👍 yes I like that

I think we should keep these names as human-parseable as possible since users might need to identify them at some point

rokroskar: so I think we can go with 

`AUTOSAVE_BRANCH_PREFIX="renku/autosave/${JUPYTERHUB_USER}"`

PR #43, opened 2019-08-08
fix: pin r-base to avoid dependency issues

rokroskar:


rokroskar:
actually I think the problem was an incompatibility between R and some system library - the RStudio on conda seems currently incompatible with R 3.6.1.

PR #51, opened 2019-08-20
modified Dockerfile to include conda installation

emmjab:


docker/r/Dockerfile
```diff
@@ -88,6 +104,8 @@ RUN pip3 install --no-cache-dir pipx && \
     pipx inject renku sentry-sdk && \
     pipx ensurepath
 
+ENV PATH=/opt/conda/bin/conda:$PATH
```
rokroskar: I think here you just want `ENV PATH=/opt/conda/bin:$PATH`

emmjab: might be a red herring:

In a Dockerfile in a project that looks like this:
```
FROM emmjab/renku-tests:latest
#FROM renku/singleuser-r:0.3.5-renku0.5.2

# Uncomment and adapt if code is to be included in the image
# COPY src /code/src

# install the python dependencies
COPY requirements.txt environment.yml /tmp/
RUN conda env update -q -f /tmp/environment.yml && \
    /opt/conda/bin/pip install -r /tmp/requirements.txt && \
    conda clean -y --all && \
    conda env export -n "root"
```
The CI/CD fails with:
```Step 1/3 : FROM emmjab/renku-tests:latest
 ---> 98772bb3e57f
Step 2/3 : COPY requirements.txt environment.yml /tmp/
 ---> Using cache
 ---> d5475087709c
Step 3/3 : RUN conda env update -q -f /tmp/environment.yml &&     /opt/conda/bin/pip install -r /tmp/requirements.txt &&     conda clean -y --all &&     conda env export -n "root"
 ---> Running in 39e75a0a7c6a
/bin/sh: 1: conda: not found
The command '/bin/sh -c conda env update -q -f /tmp/environment.yml &&     /opt/conda/bin/pip install -r /tmp/requirements.txt &&     conda clean -y --all &&     conda env export -n "root"' returned a non-zero code: 127
ERROR: Job failed: exit code 127```
even though the conda command is available when starting up a bash shell from the Docker image.

Comment (rokroskar) [reply docker/r/Dockerfile]: Yes, it's available through bash in a login shell because the path is set via `. bashrc`. But in the case of a non-login shell you get the path set in the `ENV` which in this case points to the conda executable instead of the directory.

Comment (emmjab) [reply docker/r/Dockerfile]: I should have mentioned that I made the change to the path (add `/opt/conda/bin` to path instead of `/opt/conda/bin/conda`) before my last comment -- the Docker build with the user's conda file still fails. But at least that explains why it works anyway in `.bashrc`!

File: docker/r/entrypoint.sh [hunk 2]
```diff
@@ -8,5 +8,9 @@ for var in ${VariableArray[*]}; do
     fi
 done
 
+# add a symlink to the work directory containing the project to the home
```
rokroskar: maybe this comment can simply say: "add a symlink to the project directory in /home/rstudio"

docker/r/Dockerfile
```diff
@@ -63,6 +67,27 @@ RUN chmod +x /tini
 # This allows non-root to install python libraries if required
 RUN mkdir -p ${VENV_DIR} && chown -R ${NB_USER} ${VENV_DIR}
 
+# install miniconda (https://github.com/ContinuumIO/docker-images/blob/master/miniconda3/debian/Dockerfile)
+RUN apt-get update --fix-missing && \
+    apt-get install -y wget bzip2 ca-certificates libglib2.0-0 libxext6 libsm6 libxrender1 git mercurial subversion && \
```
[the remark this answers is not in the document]
emmjab: I didn't put them all together so that it would be clear which dependencies belong to which installation -- it's hard enough to read dockerfiles... 

if you think all the apt-get installs should all be in one place, I can add the clarifications in the README -- what do you think?

docker/r/entrypoint.sh
```diff
@@ -8,5 +8,9 @@ for var in ${VariableArray[*]}; do
...
+# where Rstudio puts you when rstudio starts up.
+ln -s /work /home/rstudio
```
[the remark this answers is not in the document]
emmjab: ah -- didn't realize, thanks for pointing that out!

docker/r/Dockerfile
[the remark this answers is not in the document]
rokroskar: yeah that's true, but it's usually better to keep that all in one layer. At least you don't really need to do the update again since it's done above.

docker/r/entrypoint.sh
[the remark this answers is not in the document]
rokroskar: actually it doesn't work, that variable is only set once the notebooks starts...

PR #54, opened 2019-09-19
chore: add new renku-python version to matrix

rokroskar:


jsam:
I guess we have to publish it first and then rerun CI to merge it?

rokroskar:
it's published: https://pypi.org/project/renku/

but maybe wasn't quick enough ;)

PR #55, opened 2019-10-28
build: bump renku version matrix

rokroskar:


emmjab:
We didn't set up the automated build-to-docker yet, right?

PR #57, opened 2019-10-31
image: Cuda 10.0 Tensorflow 1.14

### LINKED ISSUE #19: create a base tensorflow image
It should be based on the cuda image and:

- [ ] install tensorflow
- [ ] install the tensorboard jupyter extension

cc @nperraud

pameladelgado:
This PR provides a Renku image for Cuda 10.0 with Tensorflow 1.14
To test:
1. Go to https://limited.renku.ch/projects/51/launchNotebook 
2. Launch a notebook from the latest commit with GPUs and 8GBs of memory
3. From the notebook open a console and execute `python test-tensorflow-gpu.py`

Expected result: the program runs without errors and that at least one GPU device is listed in the logs (similar to https://limited.renku.ch/gitlab/pameladelgado/test2/blob/master/logs-gpu-test/test-log-2gpus.txt)

Fixes partially #19 . Tensorboard comes with the image, but to be able to access it we need to do some port forwarding before running the notebook (https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks)

Update: closes #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@
 
 extensions = \
 	r \
-	cuda-9.2
+	cuda-9.2 \
+	cuda-10.0-tensorflow-1.14
 
 DOCKER_PREFIX?=renku/singleuser
 DOCKER_LABEL?=latest
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,7 @@
 
 extensions = \
 	r \
-	cuda-9.2 
-	#cuda-10.0-tensorflow-1.14
+	cuda-9.2 cuda-10.0-tensorflow-1.14
 
 DOCKER_PREFIX?=renku/singleuser
 DOCKER_LABEL?=latest
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@
 
 extensions = \
 	r \
-	cuda-9.2 \
-	cuda-10.0-tensorflow-1.14
+	cuda-9.2 
+	#cuda-10.0-tensorflow-1.14
 
 DOCKER_PREFIX?=renku/singleuser
 DOCKER_LABEL?=latest
--- a/docker/cuda-10.0-tensorflow-1.14/Dockerfile
+++ b/docker/cuda-10.0-tensorflow-1.14/Dockerfile
@@ -1,0 +1,84 @@
+ARG BASE_IMAGE=renku/singleuser:latest
+FROM $BASE_IMAGE
+
+LABEL maintainer="Swiss Data Science Center <info@datascience.ch>"
+
+## cuda base
+## from https://gitlab.com/nvidia/cuda/blob/ubuntu18.04/10.0/base/Dockerfile
+USER root
+RUN apt-get update && apt-get install -y --no-install-recommends gnupg2 curl ca-certificates && \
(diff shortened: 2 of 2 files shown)


rokroskar:
This issue might help with getting tensorboard to show up in the session: https://github.com/jupyterhub/jupyter-server-proxy/issues/41

ableuler:
> Tensorboard comes with the image, but to be able to access it we need to do some port forwarding before running the notebook

Exposing the port for tensorboard might be tricky. The jupyterlab servers are exposes through the jupyterhub proxy. I don't think that this one can be configured to forward other ports than `http`, or can it?

rokroskar:
This is what the jupyter server proxy can be used for - see here: https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html

ableuler:
> This is what the jupyter server proxy can be used for - see here: https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html

Sorry, had not seen your response before typing mine (pretty sure it didn't load automatically).

rokroskar:
@pameladelgado did you have a look at https://github.com/lspvic/jupyter_tensorboard ?

pameladelgado:
> @pameladelgado did you have a look at https://github.com/lspvic/jupyter_tensorboard ?

I did not try this, but again this seems to need some port forwarding for port `8888` (https://github.com/lspvic/jupyter_tensorboard#use-jupyter-tensorboard-in-docker-containers) that, if I understand correctly, should be done when the notebook is launched

pameladelgado:
oh wait. I'm just also catching up with your comments :)

rokroskar:
that's the standard jupyter port that is already proxied.

ableuler:
> @pameladelgado did you have a look at https://github.com/lspvic/jupyter_tensorboard ?

That alone didn't do the job, but together with the [corresponding extension for jupyterlab](https://github.com/chaoleili/jupyterlab_tensorboard) which relies on the above one things seem to work. The image is available on dockerhub as `ableuler/singleuser-cuda-10.0-tensorflow-1.14`.

rokroskar:
btw, this runs the image and includes an example notebook: https://limited.renku.ch/projects/18/

rokroskar:
we should probably also build a version with TF 2.0 though

pameladelgado:
> we should probably also build a version with TF 2.0 though

Yes, but I think we can do another image for that. Some of our users explicity wanted 1.14.

rokroskar:
Sure, I suppose it would be best to parametrize the tensorflow version in the `Dockerfile` if possible. Then we can include the TF versions in the build matrix.

pameladelgado:
> Sure, I suppose it would be best to parametrize the tensorflow version in the `Dockerfile` if possible. Then we can include the TF versions in the build matrix.

That would be good, we can do another PR for that and maybe do the same for the python image?

rokroskar:
Shouldn't we pin the cuda version? Now it's set to `>= 10` which is different from `10.0`.

pameladelgado:
> Shouldn't we pin the cuda version? Now it's set to `>= 10` which is different from `10.0`.

It's pinned as part of the image build, see line 16
https://github.com/SwissDataScienceCenter/renku-jupyter/pull/57/files#diff-3607aedaef5ba4bc4eaf64b4d536a139R16

rokroskar:
ah ok great, then 👍

PR #59, opened 2019-11-05
added vim to Dockerfile and symlink for R in entrypoints

### LINKED ISSUE #56: add vim to base R docker image
As an RStudio user who wants to use git via the terminal in jupyterlab (as opposed to the gui git interface provided by RStudio), it would be nice to have vim installed in the image.

emmjab:
For the R template, vim was missing, and there was no symlink to the project files from the RStudio interface. Both of these were added & tested here: https://renkulab.io/projects/1176/files

Fixes #56

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -38,7 +38,8 @@ RUN apt-get update --fix-missing && \
         libsm6 \
         libxext6 \
         libxrender1 \
-        wget && \
+        wget \
+        vim && \
     apt-get purge && \
     apt-get clean && \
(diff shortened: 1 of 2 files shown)


rokroskar:
Shouldn't the git setup be the default for all images (i.e. also the python one)?

emmjab:
@rokroskar : yeah I agree we should standardize between all the images; the way it's done in the base (python) image is that there's that other file -- the `git-config.bashrc`. I thought when we were originally making the R image we had problems with however it was done in the python proj, but I can check.

PR #63, opened 2019-11-26
Updated tags travis deploy

pameladelgado:
After adding 
```
branches:
  only:
    - master
```
travis is not accepting requests of tags for deployment.
The current configuration is equivalent to https://github.com/SwissDataScienceCenter/renku-graph/blob/master/.travis.yml which works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,14 +51,15 @@ script:
   - make all
 
 deploy:
-  - &script
-    provider: script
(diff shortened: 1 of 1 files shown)


rokroskar:
you need to add a line like https://github.com/SwissDataScienceCenter/renku-graph/blob/master/.travis.yml#L5 but with a regex just for the tags

PR #65, opened 2019-11-26
feat: inject environment setup into session

### LINKED ISSUE #64: Inject user preferences into the environment
# Description

Users want to be able to set up an environment configuration that moves with them. Examples of what may be in here include:
* .bashrc
* Jupyter font sizes

# Solution

Add code to the docker entrypoint that looks for a repo in gitlab that conforms to a naming convention which includes the user's name. If it is found, copy over and append the configuration files into the environment.

Document the process in the renku documentation.

rokroskar:
This PR adds an entrypoint to the base image that copies the contents of the files found in the user's `renku-env` repository in the gitlab instance. It takes the oauth token from the environment and uses it to clone the private `renku-env` repo. If the repo isn't there it fails silently. The contents of all the files in the `renku-env` repository are _appended_ to the corresponding files in `$HOME`.

This requires the user to set up a private `renku-env` repository and it should be documented in the main docs. 

To test, you may either launch an environment from https://renkulab.io/projects/1464/ or use the image `rrrrrok/singleuser:entrypoint` in your own project. 

closes #64

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -36,6 +36,11 @@ RUN ln -s /usr/lib/x86_64-linux-musl/libc.so /lib/libc.musl-x86_64.so.1
 # install git-lfs
 RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash && sudo apt-get install git-lfs=2.8.0
 
+# Add Tini
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
+
 # switch to the notebook user
 USER $NB_USER
 
@@ -70,5 +75,8 @@ USER root
 RUN echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \
     rm -rf /wheels
 USER $NB_USER
+
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT [ "/tini", "--", "/entrypoint.sh" ]
 
 CMD '/usr/local/bin/start-singleuser.sh'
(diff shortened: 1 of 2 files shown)


emmjab:
Just to be clear -- this is intended for things like settings in `.bashrc` (i.e. emacs config?)

emmjab:
> The contents of all the files in the renku-env repository are appended to the corresponding files in $HOME.

The way I'm reading this -- `renku-env` should *just have env related files* (i.e. most of the default template stuff should be cleared out of the repo if it's created via renku)?

emmjab:
I tested in the following way & functionality seems to work:
1. Created private `renku-env` project, added a `.bashrc` with one line of contents:
```
alias helloworld="echo 'hello world'"
```
2. Created public `use-renku-env` project, updated the `FROM` line in the Dockerfile to: 
```
FROM rrrrrok/singleuser:entrypoint
```
3. Started a jupyterlab from `use-renku-env`, opened a terminal:
```
$ helloworld
hello world
```

rokroskar:
@emmjab thanks for having a look - yes, the `renku-env` project _should not be a renku project_ - alternatively, we could have a template for this? Not sure how the UI wants to handle it, one suggestion was to offer some UI elements to standardize the process.

rokroskar:
> I think the way it should work is to copy over, recursively, any folders that are defined in the repository. The appending to existing file behavior works well for certain files (like .bashrc), but might no work well for everything. I think we need to clarify that aspect and maybe make it a bit more precise (e.g., maybe it should only apply to bash dot files and a different way of resolving conflicts should apply elsewhere).

It's hard to see how to make some generic behavior to support all use cases. One option that comes to mind is to use something like the `manifest.yml` file we have in the templates. It could have sections to define specific behaviors. Something like:

```
copy: 
  - name: scripts
     dst: ${HOME}/my_scripts
append: 
  - .bashrc
```

This would copy `scripts` to `${HOME}/my_scripts` and append the contents of `.bashrc` to the existing one. 

The other thing that would be nice is to exclude the `renku_env` repo from the project listing in the UI. 
This would copy

PR #66, opened 2019-11-27
pinned jupyterlab version and added pip timeout

emmjab:
Hypothesizing why some of the matrix build is failing for these images.

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -75,7 +75,7 @@ ARG JUPYTERHUB_VERSION=0.9.6
 RUN conda install --quiet --yes \
     'notebook=6.0.0' \
     "jupyterhub=${JUPYTERHUB_VERSION}" \
-    'jupyterlab=1.0.9' \
+    'jupyterlab=1.0.0' \
     'nodejs>=6.11.5' \
     conda-build && \
     conda build purge-all  && \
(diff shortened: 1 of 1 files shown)


docker/r/Dockerfile
```diff
@@ -87,7 +87,7 @@ RUN jupyter notebook --generate-config && \
     rm -rf /home/$NB_USER/.cache/yarn && \
     pip install -U --no-cache-dir pip wheel 
 
-RUN pip install --no-cache-dir \
+RUN pip install --no-cache-dir --default-timeout=50 \
         jupyter-rsession-proxy \
         jupyterlab==1.0.0 \
```
rokroskar: then maybe remove this line?

emmjab: just added that line. the default is 15.

rokroskar: I meant the `jupyterlab==1.0.0` part since it is set above in the conda install

emmjab: oh -- yeah. also the build failed because conda doesn't have that version ha.

rokroskar: maybe that's why I pinned it with `pip` :)

rokroskar: but you might as well take this opportunity to check if bumping the version works with all the extensions we use

PR #69, opened 2020-02-03
updated syntax for pipx

emmjab:
latest version of pipx (15 from 14) has new syntax

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -62,10 +62,8 @@ RUN conda install gxx_linux-64
 ENV RENKU_DISABLE_VERSION_CHECK 1
 
 RUN python3 -m pip install pipx
-RUN pipx --version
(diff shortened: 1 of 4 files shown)


rokroskar:
So `pipx` I guess changed other parts of the interface as well... :/

docker/base/Dockerfile
```diff
@@ -63,7 +63,7 @@ ENV RENKU_DISABLE_VERSION_CHECK 1
 
 RUN python3 -m pip install pipx
```
rokroskar:
```suggestion
RUN python3 -m pip install pipx>=0.15.0.0
```

emmjab:
> Please add the `pipx` version requirement wherever we install `pipx` just to make it clear.

Done. I think I clicked too many buttons, the latest commit & check add the versions and the tests passed.

PR #72, opened 2020-02-14
build: create pr_base_docker.yml

emmjab:
small test for github actions based on template


.github/workflows/build_and_push.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Renku Docker pull request images for testing
+      if: github.event_name == 'pull_request'
+      run: |
+        docker build docker/${{ matrix.EXTENSIONS }} --tag $RENKU_DOCKER_TAG
+        docker push $RENKU_DOCKER_TAG
+      env:
+        GIT_HASH: $(git rev-parse --short "$GITHUB_SHA")
```
rokroskar: There is a syntax in github actions you can use to set environment variables from a step that will be available to other steps. So you can have a step before this one that would do something like what is done here: https://github.com/SwissDataScienceCenter/renku/blob/ed178ffd84ff58ed9cc1a5e7c79d62e958ae2c09/.github/workflows/pull-request-test.yml#L52

.github/workflows/build_and_push_pr.yml
```diff
@@ -0,0 +1,33 @@
+name: Renku Base Docker Image CI
+
+on: [push]
```
rokroskar: This is also going to run on master and tags - is that what you want?

.github/workflows/build_and_push_tags.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Docker Login
+      uses: Azure/docker-login@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - uses: actions/checkout@v2
+    - name: Build the Renku Docker pull request images for testing
+      run: |
+        export RENKU_TAG_NAME=$(echo $GITHUB_REF | cut -d / -f 3)
```
rokroskar: If the only difference is the tag name why not have this as a conditional step at the top and export it as an environment variable? Then you only need one file/workflow and can reuse almost all of the code.

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        if [[ ${{ github.ref }} == refs/tags* ]]; then
+          export LABEL=$(echo ${{ github.ref }} | cut -d / -f 3)
+        else
+          export LABEL=$(echo ${{ github.sha }} | cut -c 1-7)
+        fi
+        export DOCKER_NAME="renku/singleuser-$EXTENSION"
+        export DOCKER_TAG="$LABEL-renku$RENKU_VERSION"
+        docker build docker/$EXTENSION --tag $DOCKER_NAME:$DOCKER_TAG
+        docker push $DOCKER_NAME:$DOCKER_TAG
```
rokroskar: Push only on master and tags

PR #75, opened 2020-02-28
build: use powerline for terminal prompt

### LINKED ISSUE #31: simplify command line
Some pod names can be very long which results in a very long command-line in the jupyterlab terminal. It should be shortened -- additionally, a tool like bash-it could be installed to make it a bit nicer to work with git in the shell.

rokroskar:
Prettifies the terminal shell with powerline. 

screenshot: 

image

closes #31

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -61,14 +61,28 @@ RUN conda install gxx_linux-64
 # install renku-python
 ENV RENKU_DISABLE_VERSION_CHECK 1
 
-RUN python3 -m pip install pipx>=0.15.0.0
+RUN python3 -m pip install "pipx>=0.15.0.0"
 ARG RENKU_PIP_SPEC="renku"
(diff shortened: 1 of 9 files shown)


rokroskar:
Thanks for testing @emmjab - btw, do you have powerline fonts installed on your computer already?

rokroskar:
I resorted to using the "compatible" mode that results in an uglier font but the alternative is requiring that the user installs fonts on their system, which I think is a no-go.

rokroskar:
I just changed all the image builds to use compatible mode and to not install any fonts. Please re-approve.

emmjab:
> Thanks for testing @emmjab - btw, do you have powerline fonts installed on your computer already?

Nope!

> I just changed all the image builds to use compatible mode and to not install any fonts. Please re-approve.

Re-testing now.

Also I agree it would be nice to have a `common/` dir for dependencies that all the images share, in a following PR; so long as we make it clear what's going on in the docker file.

PR #76, opened 2020-03-03
build: update jupyterhub version to 1.2

lorenzo-cavazzi:
After SwissDataScienceCenter/renku-notebooks#253 is merged, it makes sense to update the `jupyterhub` library to version 1.1 . Since the image `1.1` is not available, it seems that the `1.2` is the closest


rokroskar:
I don't think it exists yet?

lorenzo-cavazzi:
Oh well, I tried only on the base image where `JUPYTERHUB_VERSION` is acually used for `jupyterhub/singleuser` where `1.2` exists but not `1.1` (ref: https://hub.docker.com/r/jupyterhub/singleuser/tags).
I see now that on `r` based images that values is used as the python `jupyterhub` package version. In that case, it should be `1.1.0` (ref: https://pypi.org/project/jupyterhub/#history)

lorenzo-cavazzi:
Image preview available (only for python 3.7) by using this docker immage
`FROM lorenzocavazzitech/singleuser:200304jh1`

Or create a new python project from https://lorenzo.dev.renku.ch/

pameladelgado:
can we maybe unify the jupyterhub version accross dockerfiles and makefile? 
we have 1.2 in some cases and 1.1 in some other cases, or is there a reason for this?

lorenzo-cavazzi:
`JUPYTERHUB_VERSION` is used for defining the docker image version in the python image and for pinning the python library in the other images. I agree this should be unified -- I would personally remove the variable at all and leave the numbers. Should we address that in another issue/PR?

pameladelgado:
> Should we address that in another issue/PR?

yes, this is fine by me

rokroskar:
why do today what you can do tomorrow! :)

lorenzo-cavazzi:
Images for testing purposes

```
lorenzocavazzitech/renkulab-image-python-20200305
lorenzocavazzitech/renkulab-image-r-20200305
lorenzocavazzitech/renkulab-image-bioc-20200305
lorenzocavazzitech/renkulab-image-cuda10-20200305
```

pameladelgado:
tested the cuda image provided by @lorenzo-cavazzi and it seems to work fine, thanks!

rokroskar:
I get this again: 

image

@emmjab any clue what fixed it last time?

rokroskar:
hmm looks like I can't modify the git config:

```
error: could not lock config file /home/jovyan/.gitconfig: Permission denied
error: could not lock config file /home/jovyan/.gitconfig: Permission denied
```

emmjab:
> @emmjab any clue what fixed it last time?
yeah... i think it was the version of the jupyterlab-git extension

lorenzo-cavazzi:
We can pin that down, but I am unable to reproduce that problem.
Tried on both a new project and an old one where I replaced the Docker image

https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-python-0-6-again
https://dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial

rokroskar:
I tested the python image - seems to work fine

lorenzo-cavazzi:
Should we move on with these? Or is there something to fix/change?

rokroskar:
Ah yes sorry, I kind of lost track - one of the issues was that the new rstudio images had a broken terminal. But that's not really a reason to stall this PR.

lorenzo-cavazzi:
Ok great!
I guess only R images need to be tested yet, right? Pinging @emmjab about this :slightly_smiling_face:

rokroskar:
There is a new version of r-session-proxy - trying it out right now

emmjab:
new version of jupyter-rsession-proxy pinned in this PR (which is failing to build for some unrelated reason): https://github.com/SwissDataScienceCenter/renkulab-docker/pull/88

docker/r3.6.1/fix-permissions-100.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -e
+
+for d in "$@"; do
+  find "$d" \
+    ! \( \
+      -group 100 \
+      -a -perm -g+rwX  \
+    \) \
+    -exec chgrp 100 {} \; \
```
rokroskar: Maybe make this use the `NB_GROUP` environment variable (or is it`NB_GID`)? instead of having two different versions of the same script.

emmjab: NB_GID, and from all my testing, if you set this to 100 to fix the conda directories, then NB_USER can't the conda packages, so it has to be 1000. But, if you set them to 1000, then the user who's running in the notebook doesn't have access, it needs them to be 100.
🤷‍♀

rokroskar:
This is where the new changes are testable: https://rok.dev.renku.ch/projects/rokroskar/is-everything-broken - all devs should have access

rokroskar:
The two latest commits use R and python respectively

rokroskar:
if you want to quickly add the matrix change I can re-approve

lorenzo-cavazzi:
A bit late to the party, but I also tested and everything seems to work fine :tada:

PR #77, opened 2020-03-03
build: change docker image names

emmjab:
This change improves the semantics of the docker image names to accommodate specifications in the images of:
* the version of renku-python
* the name and version of the language or base image
* a "label" identifier for the version of image's other dependencies & features


rokroskar:
I made all the current checks required

PR #80, opened 2020-03-04
docs: update the readme to current image names

emmjab:


lorenzo-cavazzi:
ops sorry I see this late, otherwise I would probably not have pushed #82 

Anyway, I would leave a convenient `make all` command and some sort of `make base` -- could also be an example like `you can use make <name> to build a template, e.g. make py3.7`

emmjab:
I think #82 is fine if you add the naming changes as well :) 
maybe we can merge this and you can add the section back w/the updated make syntax in your PR?
thoughts?

lorenzo-cavazzi:
I thought we could leave the `make all` part directly, but it's ok also to re-add it in the other PR :slightly_frowning_face:

PR #82, opened 2020-03-04
fix: restore makefile commands

### LINKED ISSUE #81: Fix Makefile or Readme
After #77 has been merged, the commands `make base` and `make all` don't work anymore.
We should update the `Makefile`. Should we also change `make base` to something else, like `make python`?

lorenzo-cavazzi:
fix #81


Makefile
```diff
@@ -17,8 +17,8 @@
 # limitations under the License.
 
 extensions = \
-	r-bioconductor_3_10 r \
-	cuda-9.2 cuda-10.0-tensorflow-1.14
+	bioc3_10 r3.6.1 \
+	cuda9.2 cuda10.0-tf1.14
 
 DOCKER_PREFIX?=renku/singleuser
```
emmjab: we changed the prefixes as well:

`renku/renkulab`

where `renku/renkulab-renku[version]-[languageVersion]-[label]`
and label is the commit hash OR version-tag from the repo

emmjab:
Questions:

1. So the CI no longer uses the Makefile (when we switched from travis to github actions), but I get that it's still useful to keep the makefile for local builds & tests.

2. The naming conventions on the variables has been updated a bunch in the github actions, and in order to keep this maintainable it might make sense to update them in the makefile as well.

3. Is there a way to run the github actions locally instead?

4. the multi-step builds haven't been fixed yet in the github actions (i.e. the 2 cuda builds that rely on "base" which is now py3.7); mostly because I can't see any workaround for duplicated code (and the Makefile *also* uses the duplicated codeblock, so maybe that's the way to go...).

PR #88, opened 2020-03-24
build: pin rsession and re-add powerline

emmjab:
* pinning rsession-proxy for R images to make sure the terminal works
* adding back powerline to R images

This PR also intends to fix the ability for R users to install packages, but at the current state it does not do this.


rokroskar:
We can close this right?

emmjab:
CLOSING! All of the changes that needed to be made in here were made in https://github.com/SwissDataScienceCenter/renkulab-docker/pull/76 instead.

PR #94, opened 2020-04-06
Add nbresuse for resource display in notebooks

### LINKED ISSUE #33: Add notebook resource usage extension
Install https://github.com/yuvipanda/nbresuse in the base image.

emmjab:
closes #33 

Adds https://github.com/jtpio/jupyterlab-system-monitor which uses https://github.com/yuvipanda/nbresuse.

I think we should add some configs in `~/.jupyter/jupyter_notebook_config.py` like the example before merging:
```
c = get_config()

# memory
c.NotebookApp.ResourceUseDisplay.mem_limit = <size_in_GB> *1024*1024*1024

# cpu
c.NotebookApp.ResourceUseDisplay.track_cpu_percent = True
c.NotebookApp.ResourceUseDisplay.cpu_limit = <number_of_cpus>
```

### DIFF
--- a/docker/py3.7/Dockerfile
+++ b/docker/py3.7/Dockerfile
@@ -55,6 +55,7 @@ RUN python3 -m pip install -U pip && \
     jupyter labextension update @jupyterlab/hub-extension --no-build && \
     jupyter labextension install @jupyterlab/git --no-build && \
     jupyter labextension install @jupyterlab/toc --no-build && \
+    jupyter labextension install jupyterlab-topbar-extension jupyterlab-system-monitor && \
     jupyter labextension list && \
     pip install jupyterlab-git && \
     jupyter serverextension enable --py jupyterlab_git && \
--- a/docker/py3.7/requirements.txt
+++ b/docker/py3.7/requirements.txt
@@ -2,3 +2,4 @@ papermill<1.2.0
 requests>=2.20.0
 jupyterlab==1.2.6
 jupyterhub==1.1.0
+nbresuse==0.3.3


emmjab:
Can test with this: `FROM renku/renkulab:renku-py3.7-3256b1f`

rokroskar:
hmm now we have both? nbresuse in the lower left and jupyterlab-system-monitor in the upper right? 😄

rokroskar:
Are we already setting anything in `~/.jupyter/jupyter_notebook_config.py`? I don't recall doing anything with it before.

emmjab:
> hmm now we have both? nbresuse in the lower left and jupyterlab-system-monitor in the upper right?

it was always in both places, just now the one in the top right is more colorful

> Are we already setting anything in ~/.jupyter/jupyter_notebook_config.py? I don't recall doing anything with it before.

No I don't think we are -- but since the purpose of the configs is to give visual indicators of like percentage usage of resources, I think it's of limited value if we don't indicate what the thresholds are? When I tested I didn't see the colorful CPU thing being measured, and I assumed it was because we didn't set a threshold.

Furthermore: I watched the memory bar change when I `renku dataset import`ed like 30GB or something of data, but not very much at all. Does someone already have a process that actually uses a lot of memory? If anyone has one we should turn it into a `renku-qa` project also.

I don't really know what kinds of things this is measuring / what will be helped by having it, so it's hard for me to test. Can whoever tests explain this :D

rokroskar:
It gets the threshold from the environment variables set automatically by jupyterhub. If you run it on a renku deployment you'll see it. But I don't see the cpu - that would be useful to add.

You can 

```
import numpy
x = np.arange(<large-integer>)
```

and watch the memory go up. :)

emmjab:
Thanks! `numpy` ftw.

I watched the mem bar go up, but I think one of the options is refresh rate, for instance. In this case with:

```
import numpy as np
x = np.arange(10e10)
```

It went orange suddenly, and then the kernel died 😁 

Secondly, maybe I'm wrong, but I think `NotebookApp.ResourceUseDisplay`options below are specific to this app? None of these parameters are set -- I grepped the `~/.jupyter/jupyter_notebook_config.py` for these params (provided by the docs) and they're not there; also checked `env` for similar looking names:

```
c = get_config()

# memory
c.NotebookApp.ResourceUseDisplay.mem_limit = <size_in_GB> *1024*1024*1024

# cpu
c.NotebookApp.ResourceUseDisplay.track_cpu_percent = True
c.NotebookApp.ResourceUseDisplay.cpu_limit = <number_of_cpus>
```

There's also the option to set these values when spawning the notebook.

rokroskar:
Yes those must be for this extension specifically - we could add them in the image build. I would just add `c.NotebookApp.ResourceUseDisplay.track_cpu_percent = True`.

emmjab:
just checking before I do this -- that would involve:

1. copying the default `jupyter_notebook_config.py` file to this repo (where all the settings are commented out, but would be nice to know what they are so I would leave them in the file...),
2. adding `c.NotebookApp.ResourceUseDisplay.track_cpu_percent = True` for this plugin to that file, and then,
2. `COPY jupyter_notebook_config.py ~/.jupyter/jupyter_notebook_config.py` in each Dockerfile.

seems like a lot of overhead, and the defaults in the file will get outdated over time (sure they're commented out, but something to remember...).

rokroskar:
We don't _have_ to have the cpu monitor - the memory usage will get reported correctly without any changes to the notebook config options so we can also leave it out for now.

emmjab:
Ok my preference is to leave out the CPU monitor for now, and add it in after standardizing the images (where the lines I pasted won't have to be copied around everywhere).

rokroskar:
agreed!

PR #95, opened 2020-04-06
Auwerxlab/build multiple rversions

emmjab:
* build multiple R versions including newest latest (3.6.3)
* build multiple bioc versions (including devel)

---

originally submitted as #90 by @alexisrapin


.github/workflows/build_and_push_to_docker.yml
```diff
@@ -30,23 +28,128 @@ jobs:
         password: ${{ secrets.DOCKER_PASSWORD }}
 
     - uses: actions/checkout@v2
-    - name: Build renku project docker images
+    - name: Build renku project python+ docker images
```
rokroskar, line 31: Shouldn't this job be split into two steps? First building the python base images and then the images that depend on them? Right now the cuda images are not built correctly at all. It should be something like

```
jobs:
  base-python:
    <build base python using the renku-version matrix>
  base-extensions:
    depends-on: base-python
    <build extensions, passing the correct python base image using `--build-arg` to `docker build`>
```

emmjab: Yeah I couldn't figure that out -- tripped up by the matrix.

I'll try out your suggestion, thanks!

emmjab: in terms of building the cuda images correctly and being able to specify a latest as default that relies on the right image for the cuda images, I think we should also move to the naming convention already in the makefiles so that this makes sense: `renku/renkulab-py3.7:latest`

and then we also have:
`renku/renkulab-bioc:3_10-renku0.10.2-hash`
`renku/renkulab-r:3.6.3-renku0.10.2-hash`
...
etc.

emmjab: i didn't make the naming changes for now, just pointed at the current latest for default (and passed in the bulid-arg so the default will never be used anyway in the action)

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -30,23 +72,129 @@ jobs:
         password: ${{ secrets.DOCKER_PASSWORD }}
 
     - uses: actions/checkout@v2
-    - name: Build renku project docker images
+    - name: Build renku project python-based docker images
```
rokroskar, line 75:
```suggestion
    - name: Build renku project python-based extension images
```

Maybe?

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -30,23 +72,129 @@ jobs:
...
       run: |
         if [[ ${{ github.ref }} == refs/tags* ]]; then
           export LABEL=$(echo ${{ github.ref }} | cut -d / -f 3)
         else
           export LABEL=$(echo ${{ github.sha }} | cut -c 1-7)
         fi
+        if [ -z $RENKU_VERSION ]; then
+          export RENKU_PIP_SPEC=renku
```
rokroskar, line 83: It's not necessary to set `RENKU_PIP_SPEC` because it's just taking whatever was set in the base image. See the warning here: https://github.com/SwissDataScienceCenter/renkulab-docker/runs/570365713?check_suite_focus=true#step:4:430

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -30,23 +72,129 @@ jobs:
...
+        else
+          export RENKU_PIP_SPEC="renku==${RENKU_VERSION}"
+        fi
+
         export DOCKER_TAG="renku$RENKU_VERSION-$EXTENSION-$LABEL"
+        docker build docker/$EXTENSION \
+          --build-arg BASE_IMAGE="$DOCKER_NAME:renku$RENKU_VERSION-py3.7-$LABEL" \
+          --build-arg RENKU_PIP_SPEC=${RENKU_PIP_SPEC} \
```
rokroskar, line 91: Same as above - `RENKU_PIP_SPEC` is not needed

.github/workflows/build_and_push_to_docker.yml
emmjab: good point; changed to: `Build renku project python-based docker image extensions`

PR #104, opened 2020-04-28
update base to use jupyterlab 2.0

### LINKED ISSUE #68: Fix symbolic link from home directory to repo
When launching a shell in an interactive session, the user is dropped into the checked-out repo at `/work/<repo-name>/`. The users home directory `/home/jovyan` contains a `work` folder which is empty. I assume that this work folder is supposed to be a symbolic link to `/work`, but it is not. 

Alternatively, one could create a symbolic link directly to the repo, `/home/jovyan/<repo-name> --> /work/<repo-name>`.

### LINKED ISSUE #93: Standardize images
We're up to 6 different images with 3 different bases:

- jupyterhub/singleuser (with py3.7, cuda 9 and cuda 10, and now julia)
- rocker/rstudio (with 3.6.1 R)
- bioconductor/bioconductor_docker (version 3_10, and which uses rocker/rstudio as a base)

Currently, each image "flavor" (i.e. language version [e.g. py3.7], docker base [e.g. bioconductor], specific dependency [e.g. cuda]) has its own directory, which has its own Dockerfile and other dependencies.

Why this is good:
All files required for building the image are together in one descriptively named directory, so it's easy to know and change what's in the image.

Why this is bad:
We want to maintain some semblance of consistency between what we offer in the different images, and maintaining this is tedious. In the best case it just involves copying the same code 6 times, but in less good cases installed versions of software dependencies become inconsistent, e.g. when a new feature is added to one type of image and not the others.

Why this is a hard problem:
- Each of the different bases come with different installed software and (potentially) different (potentially pinned) versions.
- Different types of images do have different features and we'll want to keep those separate
- Even when you make the same change across different images, you might not get the same results.

Related issues:
- Set up automated testing: [insert link to ticket]
- Grouping CI based on base images (maybe only rebuild images that changed) [insert link to ticket]

### LINKED ISSUE #100: base images on the jupyter/base-notebook
The only thing the singleuser image accomplishes is that it installs jupyterhub of a particular version. We do that anyway through the `requirements.txt` and can therefore just base our image off of `jupyter/base-notebook` directly.

rokroskar:
This change uses a multi-stage build to use the conda artifacts, renku installation, and user environment setup files from the `py3.7` image directly in the `R` and `bioc` images. This way we minimize the need for keeping track of the python/jupyter/renku dependencies and the environment set up (powerlien shell, PATH configuration etc.) in multiple places. 

The builds have to wait for the python stage to finish first, but because there is no more installation of python/jupyter components happening any more in the R images, the total CI time is about the same as before. 

These changes together mean this closes #93.

This PR also closes #68.

Also closes #100 

remaining TO-DO: make the Julia image also based on the py3.7 base image.


rokroskar:
Hmm good point - I guess apart from adding latex it's more or less the same thing as what we're adding in there anyway. But we can change to that one if you think it's needed?

docker/py3.7/requirements.txt
```diff
@@ -1,5 +1,7 @@
 papermill<1.2.0
 requests>=2.20.0
-jupyterlab==1.2.6
 jupyterhub==1.1.0
 nbresuse==0.3.3
+jupyterlab-git==0.20.0rc0
```
emmjab, line 5: it looks like `jupyterlab-git` got updated to 0.20.0 on may 5, so maybe we can remove the `rc0`

docker/py3.7/Dockerfile
```diff
@@ -1,19 +1,19 @@
-# Build as renku/singleuser
 # Run with the RenkuKubeSpawner via JupyterHub
-
-ARG JUPYTERHUB_VERSION=1.2
-ARG BASE_IMAGE=jupyterhub/singleuser:$JUPYTERHUB_VERSION
+ARG BASE_IMAGE=jupyter/base-notebook:8fd7ada2a374
```
emmjab, line 2: maybe you already said, but why the one with the hash and not any of the particular tags? (also it was updated again 2 days ago so we could bump this); https://hub.docker.com/r/jupyter/base-notebook/tags

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -15,9 +15,6 @@ jobs:
           - ""
```
emmjab, line 15: for the un-versioned renku-python which corresponds to the latest `--pre`, is there an easy way to get the hash of the commit that renku is being installed from? e.g. can we call this version `dev[hash]`

Makefile
```diff
@@ -17,15 +17,14 @@
...
-	py3.7 \
 	bioc \
 	r \
 	cuda-9.2 \
 	cuda-10.0-tf-1.14
 
 DOCKER_PREFIX?=renku/renkulab
 DOCKER_LABEL?=latest
-JUPYTERHUB_VERSION?=1.2
+JUPYTERHUB_VERSION?=1.1.0
```
emmjab, line 27: the jupyterhub version is pinned in the new base, right? https://hub.docker.com/r/jupyter/base-notebook/dockerfile

docker/py3.7/Dockerfile
rokroskar: which tag would you choose? They are all the same image btw.

emmjab: Probably this one `jupyter/base-notebook:hub-1.1.0` because we care about the version of jupyterhub to be compatible with whatever we have in `notebooks`?

rokroskar: yeah but that means that everything else can change because they can push a new version that has all other libraries changed _except_ JH and we wouldn't know. If we use a specific hash we guarantee that it's always going to be exactly the same image.

rokroskar: If you want, we can run `pip freeze` in the base image and store the relevant versions of dependencies that we care about somewhere in a comment, for example

emmjab: ohhh, no that's a good point; now I understand why they have all those different tags (and also that we could but shouldn't do a similar thing). we can keep the hash!

docker/py3.7/requirements.txt
rokroskar: thanks for checking that!

.github/workflows/build_and_push_to_docker.yml
rokroskar: unfortunately not atm - we could change our versioning scheme to include tags, then they would be like `v0.10.3-19-gee950ec` (meaning 19 commits since v0.10.3, hash `ee950ec`)

emmjab: that would be cool 👍

rokroskar: I can discuss it with the other renku-python people... though we are in favor of not publishing the dev releases to pypi anyway and just installing from git as needed

emmjab: huh -- i didn't think too hard about how the dev releases were available -- https://pypi.org/project/renku/#history > there's no way of getting this from pypi? we can't use the 0.10.4.dev14?

rokroskar:
Thanks @emmjab! that is a great checklist! If only it were automated... 🤔 😄

.github/workflows/build_and_push_to_docker.yml
rokroskar: I don't see an easy way atm - but I agree that we should fix this. Maybe we can open an issue?

PR #107, opened 2020-05-08
build: add bioc 3_11

emmjab:


.github/workflows/build_and_push_to_docker.yml
```diff
@@ -137,7 +136,7 @@ jobs:
         # selected R versions available for rocker/rstudio
         RVERSION: ${{ matrix.RVERSIONS }}
 
-  build-r-dev:
+  build-r-ubuntu:
```
rokroskar, line 139: wait - why do we need `build-r` and `build-r-ubuntu`?

emmjab: because the naming changed; 3.6.3 and 4.0.0 are available with -ubuntu-18.04 and the others aren't.

if you have a better idea feel free to propose!

emmjab: it was that or add in another if clause; and i wasn't sure if we wanted to keep `rockerdev` or not; maybe we do.

rokroskar: Oh ok I see, I assumed they did the os-specific versions for all versions of R. Or is it just that those exist only on ubuntu-18.04?

emmjab: yeah just ubuntu*

emmjab: https://hub.docker.com/r/rocker/verse/tags

PR #112, opened 2020-05-29
build: change the naming to exclude renku-python version

### LINKED ISSUE #111: make image naming independent of renku-python version
In our current naming scheme, the version of renku-python is injected into the image tag. This means that the template needs to be aware of the version and any time the renku-python should be updated, the user (or admin) needs to look up the appropriate image tag to use. 

The proposal here is to make the exact version of renku-python a "soft" requirement of our base images. Since `renku` is installed via `pipx` into its own virtual environment, installing a different version is easy and has no other side effects for the runtime. The version that is installed in the base image can therefore be easily overriden - on RenkuLab this can happen through a change of an environment variable in a CI job. 

The downside is that the newest images won't necessarily contain the newest renku-python version. Though not we could adopt a convention of `latest` which would always contain the latest non-dev version of the `renku-python` package. 

The images would now look like:

`renku/renkulab-py3.7:0.6.5` (specific version of the runtime - renku-python at time of image build but unspecified in the tag)
`renku/renkulab-py3.7:latest` (latest runtime, latest renku-python)

This change will allow us also to more easily upgrade user projects to latest versions of the runtime and/or renku-python.

rokroskar:
closes #111

related https://github.com/SwissDataScienceCenter/renku-project-template/pull/86


.github/workflows/build_and_push_to_docker.yml
```diff
@@ -36,13 +31,12 @@ jobs:
...
-        export DOCKER_TAG="renku$RENKU_VERSION-$LABEL"
         docker build docker/py \
           --build-arg RENKU_PIP_SPEC=${RENKU_PIP_SPEC} \
-          --tag $DOCKER_NAME-py:3.7-$DOCKER_TAG
-        docker push $DOCKER_NAME-py:3.7-$DOCKER_TAG
+          --tag $DOCKER_NAME-py:3.7-$LABEL
+        docker push $DOCKER_NAME-py:3.7-$LABEL
       env:
-        RENKU_VERSION: ${{ matrix.RENKU_VERSION }}
+        RENKU_VERSION: "0.10.4"
```
emmjab, line 39: i thought we were taking the renku version completely out of the build, and therefore we'd also want to remove the pipx installation from these files (because I thought i saw in `renku-project-template` we were including the renku installation in the Dockerfile there?)

ableuler: I was in favour of completely taking it out, but I think the majority agreement was to always have a Renku version installed in the images.

rokroskar: I think renku should be installed - but we don't have to be extra diligent about keeping the versions in lock-step since the proper version will be installed during the build anyway.

docker/py/Dockerfile
```diff
@@ -59,7 +59,7 @@ RUN conda install gxx_linux-64 && \
 ENV RENKU_DISABLE_VERSION_CHECK 1
 
 ARG RENKU_PIP_SPEC="renku"
-RUN pipx install ${RENKU_PIP_SPEC} --pip-args="--pre --no-cache" && \
+RUN pipx install ${RENKU_PIP_SPEC} --pip-args="--no-cache" && \
```
emmjab, line 62: we might have to call this `${RENKU_PIP_SPEC}==$RENKU_VERSION` or something like this? i could be wrong but between the template install with .gitlab-ci.yml and the image install i think something isn't lining up between the name of the args and the renku version

rokroskar: `RENKU_PIP_SPEC` is supposed to be something like `renku==0.10.4` or `renku>=0.10.0` or similar. `RENKU_VERSION` would be just `0.10.4`.

.github/workflows/build_and_push_to_docker.yml
emmjab: wait i thought we agreed we don't need to specify a version at all in the build

docker/py/Dockerfile
emmjab: yeah but i think you took out the line that adds RENKU_VERSION to RENKU_PIP_SPEC

rokroskar: no - that was done in the github action before. The Dockerfile only had `RENKU_PIP_SPEC`. See https://github.com/SwissDataScienceCenter/renkulab-docker/pull/112/files#diff-7453a2391fb8e2a749f878b0fc2126acL36

We could remove this completely, but then there is no way to install a different default version in there without changing the Dockerfile so I prefer to leave it in.

emmjab: i mean you didn't put anything in here to pass in RENKU_PIP_SPEC as an arg: https://github.com/SwissDataScienceCenter/renku-project-template/pull/86/files

rokroskar: That's right - the configuration there doesn't interact with this Dockerfile, it just uses the resulting image.

PR #115, opened 2020-07-13
build: add Julia 1.5.3 (current stable) image

ciyer:
- Move Julia to version 1.5.3 (current stable)
- Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,17 @@
 [![Renku Docker Image CI](https://github.com/SwissDataScienceCenter/renkulab-docker/workflows/Renku%20Docker%20Image%20CI/badge.svg)](https://github.com/SwissDataScienceCenter/renkulab-docker/actions?query=workflow%3A%22Renku+Docker+Image+CI%22)
 
-# Renkulab Docker Images
+# RenkuLab Docker Images
 
-Renkulab Docker Images contain minimal dependencies for launching interactive
-environments like Jupyterlab and RStudio from the Renku platform. They also each
+RenkuLab Docker images contain minimal dependencies for launching interactive
+environments like JupyterLab and RStudio from the Renku platform. They also each
 contain a version of the [renku cli](https://github.com/SwissDataScienceCenter/renku-python).
 
 ## Usage
 
-The basic python (renkulab-py) and basic R (renkulab-r) images are available via
+The basic python (renkulab-py), basic R (renkulab-r), and basic Julia (renkulab-julia)
+images are available via
 [renku project templates](https://github.com/SwissDataScienceCenter/renku-project-template)
-that you select upon renku project creation on the renkulab platform, or locally
+that you select upon renku project creation on the RenkuLab platform, or locally
 via `renku init`.
 
 If you would like to use an image built from this repo that is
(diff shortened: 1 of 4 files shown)


rokroskar:
Does it not work to use the existing build with a different version of Julia?

rokroskar:
I just tried this and I had to remove the line that includes the checksum for version 1.3.1 but other than that it works fine. So I would suggest we treat it like we do with R and parametrize the builds for different versions.

ciyer:
> I just tried this and I had to remove the line that includes the checksum for version 1.3.1 but other than that it works fine. So I would suggest we treat it like we do with R and parametrize the builds for different versions. Unless of course it's not at all interesting/relevant to have images with multiple versions of Julia - in which case I agree we should just support the latest.

Since 1.3.1 is neither the current stable (that is 1.4.2) nor the long-term support release (1.0.5), I thought it didn't make sense to include it as a template for *new* projects, but I'm not embedded enough in the Julia world to know the best answer. I can try to ask during the JulaCon to see Julia users have an opinion on this.

rokroskar:
Yeah I think we should just have the latest stable for the time being. But we should then remove the `julia1.3.1` directory.

docker/julia/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG BASE_IMAGE=renku/renkulab-py:3.7-renku0.10.4-0.6.3
```
rokroskar, line 1:
```suggestion
ARG BASE_IMAGE=renku/renkulab-py:3.7-0.7.3
```

docker/julia/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
rokroskar, line 1: If it's not strictly necessary, try to avoid using a custom entrypoint. Instead, you can use the post-init.sh script like we do for setting up the R image: https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/docker/r/post-init.sh - this is executed by the entrypoint from the base image [here](https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/docker/py/entrypoint.sh#L40)

README.md
```diff
@@ -50,24 +51,25 @@ You can find these base images on
 language or base image). Read the following naming conventions below to select
 the image that's right for you:
 
-`renku/renkulab-[image flavor]:[image flavor version]-[tag|hash]`
+`renku/renkulab-[image flavor]:[image flavor version]-renku[renku python version]-[tag|hash]`
 
 For example:
-`renku/renkulab-py:3.7-0.6.4`
+`renku/renkulab-py:3.7-renku0.10.4-0.6.4`
```
rokroskar, line 57: This is actually the old tag convention...

rokroskar (changes requested):
Thanks for following up on this @ciyer! The main thing to be addressed is not overriding the entrypoint.

ciyer:
Note: https://github.com/SwissDataScienceCenter/renku-project-template/issues/102 should be resolved once the PR goes through.

rokroskar (approved):
Thanks @ciyer, glad to have this in!

Issue #118, opened 2020-03-13
Any project in an interactive env shows some issues when running renku doctor


mohammad-sdsc:
## Reproduce

1. Create a new project
2. Start an interactive env
3. Open a command line and run `renku doctor`
4. It complains that githooks are missing and project's metadata is wrong.

rokroskar:
seems to no longer be a problem - please reopen if that is not true!

PR #119, opened 2020-08-23
chore: install githooks via the entrypoint

### LINKED ISSUE #117: enable renku git hooks on project clone
The entrypoint should contain a call to `renku` to install the git hooks in the project at launch time. It's not nice to do this in the init container because it's renku version-dependent so it's best to do it in the entrypoint to keep it consistent with the renku installed in the users environment.

rokroskar:
addresses #118
closes #117

### DIFF
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -28,7 +28,7 @@ then
 fi
 
 # install git hooks
-renku githooks install
+~/.local/bin/renku githooks install || true
 
 # run the command
 $@
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -27,5 +27,8 @@ then
     ln -sf /work/${CI_PROJECT} ~/work
 fi
 
+# install git hooks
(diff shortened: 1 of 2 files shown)


ableuler (approved):
Didn't test this, but seems straightforward and makes sense.

rokroskar:
The only thing we may want to do is to make it

```
renku githooks install || true
```

so that if for some reason the user has some bizarre renku version installed that messes up this command it doesn't prevent the container from starting.

rokroskar:
Ok I've tested it now and this works.

ableuler (approved):

ableuler (commented):
Actually, don't our other images (like the Rstudio one) overwrite the entrypoint?

rokroskar:
🤦 yes, you're right... I'll make the same change to the otheres

ableuler (approved):

ableuler:
@rokroskar can this one be merged?

rokroskar:
I believe so

PR #125, opened 2020-10-14
fix: pin RStudio version

### LINKED ISSUE #124: bug: RStudio redirects to non-existent URL
In recent versions of RStudio, the `/rstudio` endpoint redirects to a non-existent URL. After a few tries it eventually brings up RStudio, but it's not clear why. The discussion [here](https://github.com/rocker-org/rocker-versioned2/issues/88) suggests that it's an RStudio version problem. The rocker stack overwrites the images with recent versions of RStudio installed so it's not obvious how to go back to a working version without overwriting RStudio explicitly in the build.

rokroskar:
closes #124

### DIFF
--- a/.github/workflows/build_and_push_to_docker.yml
+++ b/.github/workflows/build_and_push_to_docker.yml
@@ -8,8 +8,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: true
     steps:
     - name: Docker Login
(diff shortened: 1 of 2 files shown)


pameladelgado (approved):
I tested launching an environment with this image in https://staging.dev.renku.ch/projects/pamela.delgado/test-r-template and the problem is gone! :tada:
Thanks a lot @rokroskar !

PR #126, opened 2020-10-15
chore: update jupyter base

rokroskar:
Updates the base image to an updated jupyter base-notebook image which means a different version of ubuntu and python.

To verify:

- [x] python image
- [x] R image
- [x] cuda


ableuler:
python image seems to work fine!

olevski:
@rokroskar I can confirm that the fix for the cuda-tf image works.

One thing I saw is that the cuda.9.2 image (without tensorflow) is not part of the makefile. Should we fully get rid of this? Or I think it may be better to include the part that installs cuda from the cuda-tf image and make it a standalone cuda image without tf? Then the cuda-tf image can start with the cuda-only image. And the cuda-only image can be there for anyone who needs it - i.e. people who use pytorch.

rokroskar:
Thanks for checking @olevski! Good point about cuda-9.2 - afaik no one uses that anymore so we should remove it, but lets do it in a follow-up PR

rokroskar:
actually I didn't fully read what you wrote :)  Also +1 on having a cuda image without TF.

olevski (approved):

olevski:
Here is a new issue for the cuda without tensorflow image: https://github.com/SwissDataScienceCenter/renkulab-docker/issues/141

Issue #73, opened 2020-02-17
enable running of custom setup scripts at launch time


rokroskar:
In some cases it makes sense to want to have some custom script that executes when the container is started. One use-case is to `pip install -e` install a package that is developed in the project itself (it's hard to do this otherwise because the repo itself is only mounted at runtime). 

The suggestion is to have some convention for this script name like `runtime-init.sh` or something similar. 

cc @nperraud

erbou:
Another example of use cases is to prepare an environment that is user-specific and cannot be generalized to all users for the project - e.g. create site xml files required by certain packages that contain user specific attributes, such as usernames.

In that case it would make sense to let users (or groups) add custom attributes to their profiles, which can be injected in the scripts at start time.

rokroskar:
This has been addressed by using a `post-init.sh` script, but probably isn't documented well enough yet. I will open a new issue about that.

PR #130, opened 2020-10-16
fix: global git credentials and some other stuff

### LINKED ISSUE #128: unable to clone private repos
We use a git credentials helper set on the project level - these credentials should be global so that the user can interact with their other private repos as well. 

reported by @TaoSunVoyage

rokroskar:
closes #128

### DIFF
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -34,5 +34,15 @@ fi
 # install git hooks
 ~/.local/bin/renku githooks install || true
 
+# run the post-init script in the root directory (i.e. coming from the image)
+if [ -f "/post-init.sh" ]; then
+    /post-init.sh
+fi
+
+# run the post-init script in the project directory
+if [ -f "./post-init.sh" ]; then
+    ./post-init.sh
+fi
+
 # run the command
 $@
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -24,6 +24,10 @@ find /tmp/renku-env -not -path '*.git*' -type f -print0 | xargs --null -I{} sh -
 
 if [[ -v CI_PROJECT ]];
 then
+    # set the global git credentials
+    git config --global credential.helper "store --file=/work/${CI_PROJECT}/.git/credentials"
+
+    # link to the home work directory
     ln -sf /work/${CI_PROJECT} ~/work
 fi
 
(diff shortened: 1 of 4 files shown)


ableuler:
Given the name of the PR I thought I'd have to greenlight a `--global` flag somewhere, and suddenly I find myself trying to verify the entrypoint logic of our R images... This is false advertisement here 😂

rokroskar:
Yeah I had to do that to not have the same git config copy-pasted in a few places... I thought it would just fly under the radar :)

rokroskar:
To be fair, I tried to be honest about it and split it into two commits. So it could in principle be separated if you want.

rokroskar:
Oh, this also closes #73 then...

rokroskar:
I updated the title of the PR... 🥇

docker/py/entrypoint.sh
```diff
@@ -24,11 +24,25 @@ find /tmp/renku-env -not -path '*.git*' -type f -print0 | xargs --null -I{} sh -
 
 if [[ -v CI_PROJECT ]];
 then
+    # set the global git credentials
```
ableuler, line 27: This relies on what's happening in the init container. Maybe create the actual credentials here (again) and place them directly in the users home directory (ie outside of the repo)? This would make it a bit nicer to re-use the image in a non-k8s setting, but at the cost of having the same credentials in two places... In principle the init container could even remove the local credentials again after successful clone to avoid this duplication.

ableuler (approved):
Looks good and works as it's supposed to. Left one comment for something that you could consider, but I'm fine with merging the PR as is.

docker/py/entrypoint.sh
ableuler: Actually the best might be to check for local credentials here from the init container and remove them if they exist and then create the global ones.

rokroskar: I wanted to limit the places where we inject the oauth token into the environment. It would be easy for users to leak their oauth token if they e.g. copy-pasted their environment variables somewhere. They're more likely to look at their environment variables than their `.git/config` file, but I agree with you that it's not ideal the way it's implemented.

ableuler: Ok, my starting point was that I'd expect local config in the repo and global config in the users home directory, but I see your point: it's true that the environment variable of the init-container are less exposed than those of the main container. So let's merge this.

PR #132, opened 2020-11-02
fix: ensure local/bin is in PATH

rokroskar:
Fixes the problem where the jupyterlab git extension environment did not have `~/.local/bin` in the `PATH`


pameladelgado (approved):
I tested the python image and it works, thanks! :+1:

rokroskar:
Thanks for checking @pameladelgado!

PR #133, opened 2020-11-11
fix: do the renku-env after git credentials are configured

rokroskar:
Use the global git credentials for cloning the `renku-env` repo


ciyer (approved):
Tested. Works as advertised.

PR #134, opened 2020-11-13
feat: add vscode to base image

### LINKED ISSUE #83: An image that supports VSCode
Instead of JupyterLab as the UI, it might be nice to also offer VSCode as an alternative

Either within the browser:

- https://github.com/cdr/code-server

Or as a VSCode Remote Server:

- https://code.visualstudio.com/docs/remote/ssh

rokroskar:
Adds VSCode using [code-server](https://github.com/cdr/code-server) and [jupyter-vscode-proxy](https://github.com/betatim/vscode-binder).

For testing, you can ~~use this image: `renku/renkulab-py:3.7-c43683d`~~ add these lines to your `Dockerfile`:

```
USER root
RUN curl -s https://raw.githubusercontent.com/SwissDataScienceCenter/renkulab-docker/c50af197ccd7b727b9baa0aebf60602c669ab546/scripts/install-vscode.sh | bash
USER ${NB_USER}
```

closes #83

### DIFF
--- a/docker/py/Dockerfile
+++ b/docker/py/Dockerfile
@@ -29,10 +29,6 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     python-dev \
     unzip \
     vim && \
-    wget https://github.com/cdr/code-server/releases/download/v3.6.2/code-server_3.6.2_amd64.deb && \
-    dpkg -i ./code-server*.deb && \
-    rm code-server_3.6.2_amd64.deb && \
-    apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
     ln -s /usr/lib/x86_64-linux-musl/libc.so /lib/libc.musl-x86_64.so.1
 
@@ -49,12 +45,9 @@ RUN python3 -m pip install --no-cache -U pip && \
     python3 -m pip install --no-cache -r /tmp/requirements.txt && \
     jupyter labextension install --no-build @jupyterlab/toc && \
     jupyter labextension install --no-build jupyterlab-topbar-extension jupyterlab-system-monitor && \
-    jupyter serverextension enable --py jupyter_server_proxy && \
-    jupyter labextension install --no-build @jupyterlab/server-proxy && \
     jupyter lab build && \
     jupyter labextension list && \
     npm cache clean --force && \
-    code-server --install-extension ms-python.python && \
     rm -rf /home/${NB_USER}/.cache
 
 # fix https://github.com/SwissDataScienceCenter/renku-jupyter/issues/14
--- a/docker/py/Dockerfile
+++ b/docker/py/Dockerfile
@@ -29,6 +29,9 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     python-dev \
     unzip \
     vim && \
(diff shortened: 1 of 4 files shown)


docker/py/Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     python-dev \
     unzip \
     vim && \
+    wget https://github.com/cdr/code-server/releases/download/v3.6.2/code-server_3.6.2_amd64.deb && \
+    dpkg -i ./code-server*.deb && \
+    rm code-server_3.6.2_amd64.deb && \
```
olevski, line 34:
```suggestion
    rm code-server*.deb && \
```

I think you should keep the `*` in the rm command as well as the `dpkg` command. Or use the specific version in both. Did you add the `*` in case the version changes in the future?

olevski: The specific version is still hardcoded in the `wget` call btw.

olevski (commented):
@rokroskar I tested the image you mentioned. I did not go over all vs code features. But I did try launching vscode, making a new file and commiting and pushing it. That works. Also I tested out the terminal and running things - that was also fine.

I have only one comment though - should be easy to resolve.

docker/py/Dockerfile
rokroskar: good point - I actually meant to make the version a top level argument so it can be changed easily everywhere. Would that make sense you think?

olevski: Yeah I think making the image a top level argument is best.

rokroskar:
@ciyer @olevski I removed the vscode install from the base `Dockerfile` and moved the steps to `scripts/install-vscode.sh`. I invoke it with these lines, and would add this to our base template but commented out:

```
USER root
RUN curl -s https://raw.githubusercontent.com/SwissDataScienceCenter/renkulab-docker/1116ea5bfdb55bfbc50415a017fc0ca7353994b6/scripts/install-vscode.sh | bash
USER ${NB_USER}
```

Of course the exact version of the file in the raw github link will be different, but this one works while this PR is open. See https://dev.renku.ch/projects/rokroskar/test-vscode/files/blob/Dockerfile

docker/py/Dockerfile
olevski: This is not relevant anymore

rokroskar: right I'll add the argument to the template Dockerfile where it can be changed. I did make it a variable in the script.

docker/py/Dockerfile
```diff
@@ -29,7 +29,6 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     python-dev \
     unzip \
     vim && \
-    apt-get clean && \
```
olevski, line 32: @rokroskar why do you remove this command here? Isn't it good to purge the cache?

olevski (commented):
One minor comment. Everything else is fine. I tested the image created by adding the command you posted in the project's dockerfile. It works.

docker/py/Dockerfile
rokroskar: indeed, thanks for catching that... a cut-and-paste error :)

olevski (approved):

PR #139, opened 2021-02-10
chore: do the jupyter build at the end of vscode script

rokroskar:
avoid jupyter rebuilds on session launch and bump vscode version

you can test it here: https://dev.renku.ch/projects/rokroskar/test-vscode/environments/new (just select `/lab` before launching)


pameladelgado (approved):
LGTM

PR #143, opened 2021-03-29
py-vnc initial

### LINKED ISSUE #1: add VNC-enabled image
We have recently developed images that allow using a desktop via the browser for e.g. image processing applications. We should add the image build to our standard collection of base images here.

gavin-k-lee:
Based on https://renkulab.io/gitlab/lee.gavin.k/tigerxfce without any imaging applications

closes #1

### DIFF
--- a/.github/workflows/build_and_push_to_docker.yml
+++ b/.github/workflows/build_and_push_to_docker.yml
@@ -46,6 +46,7 @@ jobs:
         EXTENSIONS:
           - cuda9.2
           - cuda-tf
+          - py-vnc
 
     steps:
     - name: Docker Login
--- a/docker/py-vnc/Dockerfile
+++ b/docker/py-vnc/Dockerfile
@@ -25,8 +25,9 @@ RUN apt-get update \
     && find /var/log -type f -exec cp /dev/null \{\} \;
 
 # Get noVNC
-RUN cd /opt & \
+RUN cd /opt && \
     curl -sSfL https://github.com/novnc/noVNC/archive/v1.1.0.tar.gz | tar -zxf -
+
 
 # Patch novnc to use correct path to websockify (defaults to /)
 # Note if you use vnc.html you will need to patch ui.js to use the correct path
--- a/docker/py-vnc/Dockerfile
+++ b/docker/py-vnc/Dockerfile
@@ -1,0 +1,51 @@
+ARG BASE_IMAGE=renku/renkulab-py:latest
+FROM ${BASE_IMAGE}
+
+LABEL maintainer="Swiss Data Science Center <info@datascience.ch>"
+
+USER root
+
+RUN apt-get update \
+    && apt-get install -yq --no-install-recommends \
+        dbus-x11 \
+        firefox \
+        net-tools \
+        less \
+        xfce4 \
+        xfce4-panel \
+        xfce4-session \
+        xfce4-settings \
+        xorg \
+        xubuntu-icon-theme \
+        xterm \
+    && apt-get autoremove --purge \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /tmp/* \
+    && find /var/log -type f -exec cp /dev/null \{\} \;
+
+# Get noVNC
+RUN cd /opt & \
+    curl -sSfL https://github.com/novnc/noVNC/archive/v1.1.0.tar.gz | tar -zxf -
+
+# Patch novnc to use correct path to websockify (defaults to /)
+# Note if you use vnc.html you will need to patch ui.js to use the correct path
+# and also to override localstorage which may store an incorrect path from a
+# different session. Also resize server instead of scaling client
+RUN sed -i.bak \
+    -e "s%\('path', 'websockify'\)%'path', window.location.pathname.replace(/[^/]*$/, '').substring(1) + 'websockify'); console.log('websockify path:' + path%" \
+    -re "s%rfb.scaleViewport = .+%rfb.resizeSession = readQueryVariable('resize', true);%" \
+    /opt/noVNC-1.1.0/vnc_lite.html
+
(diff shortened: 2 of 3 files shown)


docker/py-vnc/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+# Get noVNC
+RUN cd /opt && \
+    curl -sSfL https://github.com/novnc/noVNC/archive/v1.1.0.tar.gz | tar -zxf -
+
+
+# Patch novnc to use correct path to websockify (defaults to /)
+# Note if you use vnc.html you will need to patch ui.js to use the correct path
+# and also to override localstorage which may store an incorrect path from a
+# different session. Also resize server instead of scaling client
+RUN sed -i.bak \
```
rokroskar, line 36: I wonder if having a configuration file that you copy in to the container would be more maintainable? This is pretty hard to read atm

docker/py-vnc/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+        xterm \
+    && apt-get autoremove --purge \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /tmp/* \
+    && find /var/log -type f -exec cp /dev/null \{\} \;
+
+# Get noVNC
+RUN cd /opt && \
+    curl -sSfL https://github.com/novnc/noVNC/archive/v1.1.0.tar.gz | tar -zxf -
```
rokroskar, line 29: could we parametrize the version here so that it can be reused later on when patching the config file? just to avoid errors in the future.

rokroskar (changes requested):
This looks great! One (meta)comment - I don't think we need to call this "py-vnc", it can just be "vnc". We're not going to have "R-vnc", presumably. 

Looks great otherwise, I just have a few small comments.

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         EXTENSIONS:
           - cuda9.2
           - cuda-tf
+          - py-vnc
```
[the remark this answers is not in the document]
gavin-k-lee: Sure that's fine

docker/py-vnc/Dockerfile
gavin-k-lee: Fair point. I think that's what Eric did originally. I can add that / change the set up.

docker/py-vnc/Dockerfile
rokroskar: yes, exactly

gavin-k-lee: Good idea. I'll make those changes

erbou:
This is great.

Note: vnc_renku.html is a modified version of  vnc_lite.html, sugar-coated for renku (including full screen and few other things). It still has the original noVNC copyrights. I am not sure how they'll feel to be credited for the fix - shouldn't we add a line to mention that it is derived from their work?

rokroskar:
I've tested the image now and it looks great! One question: when I open the terminal it goes to `~/Desktop` - could we make the default directory `/work/<project-repo>` like it does generally for other front-ends? If it's not an easy change don't worry - but if it's just appending a line to some configuration file, we can do this in the post-init.sh script.

gavin-k-lee:
> I've tested the image now and it looks great! One question: when I open the terminal it goes to `~/Desktop` - could we make the default directory `/work/<project-repo>` like it does generally for other front-ends? If it's not an easy change don't worry - but if it's just appending a line to some configuration file, we can do this in the post-init.sh script.

I tried quite a bit to change that! I ended up trying to put something like `echo 'cd ~/work/*' >> /.bashrc` but I wasn't successful in the end - but I didn't try the post-init.sh way.

erbou:
I think we are ok from a legal standpoint. It's just that as it is they take the credit for our changes (which I don't mind), and for our bugs (which they may mind) - I just want to point that out, but I am not sure if we need to do anything.

rokroskar:
>  I ended up trying to put something like echo 'cd ~/work/*' >> /.bashrc`

oh I think what you would need is

```
echo "cd /work/${CI_PROJECT}" >> ~/.bashrc
```

You have access to that variable once the container starts so the place to put this would be the `/post-init.sh` script

rokroskar:
That seems to work:

image

rokroskar:
Also, I tried to get into an infinite loop by opening a jupyter lab session in the desktop and then clicking on the VNC button... but luckily it didn't work :)

gavin-k-lee:
Nice! By the way how are you testing the image after it has been built? I.e. how do you spin up a jupyterlab session locally? Or is there a way to do this on renkulab?

rokroskar:
I just used the image that was built by CI in my project on dev. Then I ran `jupyter lab` in the terminal and connected to it.

gavin-k-lee:
Do we also need to update the Makefile? https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/Makefile 
Or is that old now?

docker/vnc/post-init.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
rokroskar, line 1: I believe you'll need to copy this into the image in the Dockerfile

gavin-k-lee: Gotcha

rokroskar:
> Do we also need to update the Makefile?

It's not really used anywhere, though sometimes it is nice to have. You could add the vnc section if you want it to be complete.

rokroskar:
There is one tiny issue left - the powerline prompt is missing some fonts:

image

should be:

image

But I think we should merge this and solve that in a follow-up issue.

rokroskar (approved):
🚀 Thanks @gavin-k-lee this is really great!

PR #145, opened 2021-04-08
feat: update fonts; add gnome-terminal, git-gui, gitk and their desktop icons

### LINKED ISSUE #144: VNC terminal the powerline prompt is missing some fonts
In the VNC image, the powerline prompt is missing some fonts:

image

should be:

image

_Originally posted by @rokroskar in https://github.com/SwissDataScienceCenter/renkulab-docker/issues/143#issuecomment-814121502_

gavin-k-lee:
closes #144


docker/vnc/Dockerfile
```diff
@@ -17,7 +17,10 @@ RUN apt-get update \
         xfce4-settings \
         xorg \
         xubuntu-icon-theme \
-        xterm \
+        gnome-terminal \
+        ttf-dejavu \
+        git-gui \
```
rokroskar, line 22: we may want to add `gitk` here otherwise the visualization of the git history doesn't work

gavin-k-lee: good point

rokroskar: well I've never heard of or used `gitk` but it's an option in the git gui to visualize the history so this pretty much guarantees that a user will ask about it :)

rokroskar (approved):
🚀 Thanks!!

PR #146, opened 2021-04-13
feat: add renku wallpaper and set it as the default

gavin-k-lee:
adds a renku branded wallpaper to the vnc


docker/vnc/Dockerfile
```diff
@@ -58,6 +58,13 @@ COPY gitk.desktop /home/jovyan/Desktop/
 COPY .git_icon.png /home/jovyan/Desktop/
 RUN chmod +x /home/jovyan/Desktop/gitk.desktop
 
+#################################################################
+# Add wallpaper and force the default to point to the renku one
+
+RUN mkdir -p /usr/share/backgrouns/renku/
+COPY renku_background_dots.png /usr/share/backgrounds/renku/
+RUN ln -s -f /usr/share/backgrounds/renku/renku_background_dots.png /usr/share/backgrounds/xfce/xfce-stripes.png
```
rokroskar, line 66: that's dirty! :)

docker/vnc/Dockerfile
```diff
@@ -58,6 +58,13 @@ COPY gitk.desktop /home/jovyan/Desktop/
 COPY .git_icon.png /home/jovyan/Desktop/
 RUN chmod +x /home/jovyan/Desktop/gitk.desktop
 
+#################################################################
+# Add wallpaper and force the default to point to the renku one
+
+RUN mkdir -p /usr/share/backgrouns/renku/
```
rokroskar, line 64:
```suggestion
RUN mkdir -p /usr/share/backgrounds/renku/
```

rokroskar (changes requested):

docker/vnc/Dockerfile
gavin-k-lee: Yeah I know... 

The thing is the local `~/.config/xfce4/xfconf/xfce-perchannel-xml/xfce4-desktop.xml` profile (where we can replace the stripes.png with the renku.png) isn't generated until the VNC button is clicked.

The alternative is to do it 'cleanly' in the `post-init.sh` script, where we spin up the VNC and then in that `xfce4-desktop.xml` file do a `sed` to replace it properly. 

What do you think?

rokroskar: I think this is fine - it just might be hard to find the image in case someone goes looking for it

gavin-k-lee: True. If I try to go into the desktop settings from the desktop, I get shown 

image

so there's hopefully not really any user error, unless as you say they go looking for that specific wallpaper.

rokroskar (approved):
Thanks @gavin-k-lee this is starting to look great!

rokroskar:
btw you should update the screenshot in the discourse post ;)

gavin-k-lee:
Yep that was the next thing on my to-dos

rokroskar (approved):

PR #149, opened 2021-05-03
chore: update julia version to 1.6.1

gavin-k-lee:
The latest version of the julia Dockerfiles is 1.5.3 but 1.6.1 is available now.


ciyer (approved):
I tested this, and it works as expected!

PR #155, opened 2021-05-13
fix(r): add start-notebook.sh script

ableuler:
Adds the possibility to launch the server through `start-notebook.sh`.


rokroskar:
I guess this is fine but if we are fixing this part I would rather not actually depend on that script at all and just run `jupyter lab` or whatever the entrypoint/command the image specifies...

ableuler:
I totally agree. `start-notebook.sh` is the default command of all the images in the jupyter docker stack.

rokroskar:
got it - I didn't realize that was the case, I thought they just ran `jupyter notebook` by default. 👍

rokroskar (approved):

Issue #151, opened 2021-05-04
branding: rename noVNC


gavin-k-lee:
The default vnc.html file has `noVNC` as the title and so tabs in browsers will show this. I think we should rebrand this to something like RENKU VNC/RENKU Desktop/Renku VNC/Renku Desktop.

Thoughts?

https://github.com/SwissDataScienceCenter/renkulab-docker/blob/20f653dc79d2970141f188e2f85f99baecde46b4/docker/vnc/vnc_renku.html#L17

(We have already 'doctored' this .html (to add full screen functionality))

erbou:
_Thoughts?_: I agree

Comments:

- I would prefer a name with _Desktop_ in it, and avoid using VNC (everywhere) - not everyone knows VNC.

- Should the favicon be changed? e.g. jupyterlab has its own, and it is useful when we need to find interactive environment tabs among many open tabs.

- Possibly another issue for vnc.html: when in full screen, the top bar serves little purpose other than providing a clue about the interactive environment. The (un)`Full Screen` button is not needed since that can be done with `ESC` (the browser will happily remind you of that), and `Ctrl-Alt-Del` is only for certain OS. Should we therefore hide the bar, and regain a bit of screen space?

gavin-k-lee:
Agree with the above. 
- The current favicon is the default SDSC one - maybe it could be a renku frog favicon?
- For the full screen - do you mean hide the bar when in full-screen mode? I think it'd still be good to show the option somewhere when still in non-full-screen mode

gavin-k-lee:
The 'full' `vnc.html` provides more features that can be made use of, e.g. clipboard and full screen. (See #152) We can test that out and customise the favicon etc.

erbou:
Agree, the clipboard feature is needed. Note that the 'full' vnc.html stays in the way when in full screen. My preference would be to have no, or a very minimalist vnc interface in that mode. Ideally I would like to hide the bar when in full-screen mode and show it in non-full screen (that was my earlier comment).

PR #156, opened 2021-05-17
feat: vnc upgrades 

### LINKED ISSUE #152: Clipboard (Ctrl+C/V) support in the VNC
The VNC is great but often we need to copy code to and from our 'regular computer', to the VNC session, and back.

This functionality is not by default available in the current set up and so we need to figure out how to enable this, or find out if it is not possible.

gavin-k-lee:
- adds clipboard feature to VNC
- removes top bar to give more screen real-estate
- removes some unneeded functionality in the 'full' noVNC
- rebrands some features to renku (favicon, HTML title)
- removes annoying 'bell' sound in vnc terminal

closes #152


gavin-k-lee:
The updated iteration looks like the following:

image

It can be spun up at https://renkulab.io/gitlab/lee.gavin.k/blank.

gavin-k-lee:
Due to the rebranding, this

closes #151

rokroskar:
for some reason the desktop size is not automatically scaled to the browser window for me:

image


Using firefox

gavin-k-lee:
Interesting... I was using Chrome and it works fine when resizing. I'll look into it.

erbou:
Nice - would it make sense to have additional `renku` features next to the clipboard features? I am thinking of a `renku save`, but there could be others (indicators of status, such as unsaved work and actions).

gavin-k-lee:
> Nice - would it make sense to have additional `renku` features next to the clipboard features? I am thinking of a `renku save`, but there could be others (indicators of status, such as unsaved work and actions).

From my understanding of the development plan, the idea is to have those buttons available from the UI, where the jupyter lab session is embedded in an iframe. However a second step could be to have buttons in that drop-down tab that are aliases for the buttons in the UI. Thoughts?

gavin-k-lee:
@rokroskar I believe that's fixed on firefox now.

rokroskar:
@gavin-k-lee @erbou I think adding those buttons to the VNC could be a possibility, but we should discuss this with the UI team and do a proper design of what the buttons should do and how.

rokroskar (approved):
verified that the sizing problem is gone - thanks!

PR #157, opened 2021-05-19
chore: upgrades rstudio to 1.3.959

### LINKED ISSUE #154: RStudio version is out of date
- The current RStudio version is pinned at 1.2.5042 but the latest is 1.4.
https://github.com/SwissDataScienceCenter/renkulab-docker/blob/7df1a7204497439b1c934248804fdacad94ab1b5/docker/r/Dockerfile#L33

- There are a few nice upgrades, for example using renv instead of packrat as the default package manager.
- The fix would need testing.

gavin-k-lee:
Upgrades RStudio to the newest working version, according to https://github.com/rocker-org/rocker-versioned2/issues/88#issuecomment-720013981

RStudio 1.4.X does not work yet, but #154 will be closed for now

closes #154

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -1,5 +1,5 @@
 # define build arguments
-ARG RENKU_BASE=renku/renkulab-py:3.8-0.7.6
+ARG RENKU_BASE=renku/renkulab-py:latest
 ARG BASE_IMAGE=rocker/verse:4.0.4
 
 # define base images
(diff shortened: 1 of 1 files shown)


docker/r/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # define build arguments
-ARG RENKU_BASE=renku/renkulab-py:latest
+ARG RENKU_BASE=renku/renkulab-py:3.8-0.7.6
```
rokroskar, line 2: btw, `latest` is there as a placeholder so we don't have to keep changing it. When the CI runs, it replaces this to use the base python image corresponding to the same commit.

gavin-k-lee: ah yep. sorry, i was testing the build locally so didn't revert back to latest.

rokroskar: all good! 👍

rokroskar (approved):

PR #162, opened 2021-05-25
docs: Update README.md

### LINKED ISSUE #159: improve the description of available versions in the README
It's not very easy to discover the available versions of images from our README - we should have a table of some sort that outlines the tags that people can look for (maybe with links to docker hub?)

gavin-k-lee:
fixes #159


rokroskar:
Thanks for this @gavin-k-lee - note that the `renku/renkulab-cuda-tf` image is what should be used for GPU work - I think we can remove mention of the other cuda images.

gavin-k-lee:
Alright. Should we then also remove the Cuda 9.2 builds from the GH action?

rokroskar:
Probably - I thought I had done that already but apparently not 🙈

gavin-k-lee:
I've removed the Cuda 9.2 references and cleaned up some of the md file

rokroskar (approved):
🚀 Thanks!!

PR #169, opened 2021-06-10
fix: replace ~ with ${HOME} in PATH

TaoSunVoyage:


docker/r/Dockerfile
```diff
@@ -19,7 +19,7 @@ ENV SHELL bash
 ENV CONDA_PATH /opt/conda
 
 # prepend conda and local/bin to PATH
-ENV PATH ~/.local/bin:${CONDA_PATH}/bin:$PATH
+ENV PATH /home/${NB_USER}/.local/bin:${CONDA_PATH}/bin:$PATH
```
rokroskar, line 22: `HOME` is already defined above so maybe just use that? 

```suggestion
ENV PATH ${HOME}/.local/bin:${CONDA_PATH}/bin:$PATH
```

rokroskar (approved):

PR #175, opened 2021-07-12
feat: add acceptance tests

olevski:
This adds acceptance tests with cypress.

So far the tests only check that:
- the top bar in the jupyterlab ui is present - i.e. the one that says File, Edit, etc
- the buttons that allow you to start a terminal or a python terminal are there
- that one can click the button to start a shell

Other checks can be easily added - I did this just to test.

In addition we should add tests for the vnc image and for rstudio. I will reach out to Gavin about this.


olevski:
The actions that build and push the images to renkus dockerhub I cancelled on purpose. I did not change those at all.

.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build all images
+        run: |
+          docker build -t py docker/py/
+          docker build -t cuda-tf --build-arg BASE_IMAGE=py docker/cuda-tf/
+          docker build -t generic --build-arg RENKU_BASE=py  docker/generic/
+          docker build -t julia --build-arg BASE_IMAGE=py docker/julia/
+          docker build -t r --build-arg RENKU_BASE=py docker/r/
```
rokroskar, line 23: we should probably make sure that the build args are consistent everywhere. Probably in another issue/PR

.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,64 @@
+name: Acceptance tests
+
+on:
+  pull_request:
+    types:
+    - opened
+    - edited
+    - synchronize
```
rokroskar, line 8: I would just leave `synchronize` and remove the rest - you don't need to rerun acceptance tests when someone edits a PR description, for example.

.github/workflows/acceptance_tests.yml
rokroskar: one more thing - we have a `Makefile` the builds these images - should that be used here for consistency?

docker/generic/Dockerfile
```diff
@@ -51,6 +51,9 @@ ENV PATH ${CONDA_PATH}/bin:$PATH
 
 # inject the renku-jupyter stack
 COPY --from=renku_base /opt/conda /opt/conda
+COPY --from=renku_base /usr/bin/tini /usr/bin/tini
```
rokroskar, line 54: should this be installed via apt above?

docker/generic/Dockerfile
```diff
@@ -51,6 +51,9 @@ ENV PATH ${CONDA_PATH}/bin:$PATH
...
+COPY --from=renku_base --chown=1000:1000 /usr/local/bin/*sh /usr/local/bin/
```
rokroskar, line 55: what is this copying?

docker/r/Dockerfile
```diff
@@ -71,6 +71,7 @@ COPY --from=renku_base --chown=rstudio:rstudio  /usr/local/bin/ /usr/local/bin/
 COPY --from=renku_base --chown=rstudio:rstudio \
     /home/jovyan/ /home/rstudio/
 COPY --from=renku_base /entrypoint.sh /entrypoint.sh
+COPY --from=renku_base /usr/bin/tini /usr/bin/tini
```
rokroskar, line 74: same as above - install via apt?

tests/README.md
```diff
@@ -0,0 +1,16 @@
+# Acceptance Tests for Renkulab Images
```
rokroskar, line 1: Should this just be added to the main README? The relative paths in the instructions are wrong if it stays in this directory

tests/cypress/integration/test.spec.js
```diff
@@ -0,0 +1,14 @@
...
+    cy.visit("http://127.0.0.1:8888/lab")
+  })
+  it('Can find launcher icons', function() {
+    cy.get('div.jp-LauncherCard', { timeout: 10000 })
+  })
+  it('Can find main menu at the top', function() {
+    cy.get('div#jp-menu-panel', { timeout: 10000 })
+  })
+  it('Can launch terminal', function() {
+    cy.get('div.jp-LauncherCard[title="Start a new terminal session"]', { timeout: 10000 }).click()
```
rokroskar, line 12: This doesn't actually verify that it's running, correct?

rokroskar (commented):
This is great @olevski! Just a few comments to clean things up here and there. I'm also wondering if it will be possible to enter commands in the terminal with this test setup?

.github/workflows/acceptance_tests.yml
olevski: I forgot about that. I will use the Makefile.

tests/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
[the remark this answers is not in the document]
olevski: Not really - I just kept it as part of the skeleton that cypress generates... I will remove it.

.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          docker build -t r --build-arg RENKU_BASE=py docker/r/
+      - name: Py Image Acceptance Tests
+        uses: cypress-io/github-action@v2
+        env:
+          TEST_IMAGE_NAME: py
+          TEST_USER_NAME: jovyan
+        with:
+          working-directory: tests
+          command: npx mocha test.js
+      - name: Cuda-tf Image Acceptance Tests
```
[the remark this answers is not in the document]
olevski: As a bit of a noob with github's ci/cd I did not know about this. It seems def like the thing to use here. I will apply this here.

tests/cypress/integration/test.spec.js
olevski: You are right it just verifies that one can click the button to start the terminal but it does not verify that the terminal started and is functional. I think it is fairly easy to issue a few commands and make sure they succeed.

.github/workflows/acceptance_tests.yml
olevski: @rokroskar I will add an issue to redo the Makefile and use it in both the workflow that publishes the images and this one. Because as it stands right now the Makefile is not used in either these tests but it is not used in the workflow that publishes either.

olevski: see #176

docker/generic/Dockerfile
olevski: This is copying the shell scripts that come in the base Jupyterlab image that are used to start the server. Amalthea needs these and they were not present before. I.e. Amalthea uses `start-notebook.sh` and the tests use the same command here.

.github/workflows/acceptance_tests.yml
rokroskar: ah that is true - I think the Makefile got out of sync at one point and we stopped using it. afaik it works again, but it never made it back into the CI. We could also do away with the Makefile, not sure what is better.

docker/generic/Dockerfile
rokroskar: those are installed as part of the jupyterhub python package. If they are needed, we should probably just install jupyterhub. otoh, I'm not sure they _are_ needed. `start-notebook.sh` basically calls `start.sh` which does some checks for user permissions etc but can only correct them if running as root - since we don't run the containers as root I'm not sure it has any specific function. It also executes some hooks which I'm sure we don't use at the moment, but maybe users would expect that functionality? Not sure.

.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,78 @@
...
+    types:
+    - synchronize
+
+jobs:
+  acceptance-tests:
+    strategy:
+      max-parallel: 1
+      matrix:
+        include:
+          - image_name: py
```
rokroskar, line 14: I think you need to build this one first outside the matrix and make it a dependency for the other steps. The others can then execute in parallel.

olevski: There is a way to use caching with an action called [docker-build-push](https://github.com/docker/build-push-action), but I could not get it to work in this case. It gets tricky when you have separate jobs but you wish to share some artifacts (i.e. the renku/py image) between them. I will look into this in the future however - I think the biggest problem was because of a bug in the action related to not recognizing the build--args I was passing in.

tests/README.md
olevski: I fixed the paths so that they are all as if the user has already navigated to the `test` directory.

I think the main readme is already long enough and all the test commands need the user to be in the `test` directory anyhow.

docker/generic/Dockerfile
olevski: Ok I remove these - now all the images just use the `jupyter lab ...` command to start.

.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download Base Image
+        uses: actions/download-artifact@v2
+        with:
+          name: py
+          path: /tmp
+      - name: Load Base Image in Docker
+        run: |
+          docker load --input /tmp/py.tar
```
rokroskar, line 52: this would never have occurred to me 👍

rokroskar (approved):
This is a long-awaited improvement for this repo - thanks!! 🚀

PR #178, opened 2021-07-16
update python version 3.8 to 3.9

gavin-k-lee:
With JupyterLab 3.0, the default Python version is now 3.9.


gavin-k-lee:
image

Confirmed with minimal and python images

rokroskar (approved):

PR #183, opened 2021-08-26
Adapt to Amalthea

ableuler:
The goal of this is to maintain as much of the current behaviour with Amalthea-based sessions, while also keeping the images backwards compatible for now.

All the images which I've touched in this PR are on dockerhub as:
- ableuler/renkulab-py:b02a1a2
- ableuler/renkulab-r:b02a1a2-r4.0.4
- ableuler/renkulab-vnc:b02a1a2
- ableuler/renkulab-julia:b02a1a2

Projects on dev using these images are available under 
- https://dev.renku.ch/projects/andi/amalthea-python-image
- https://dev.renku.ch/projects/andi/amalthea-r-image
- https://dev.renku.ch/projects/andi/amalthea-vnc-image
- https://dev.renku.ch/projects/andi/amalthea-julia-image

as well as the respective path on the CI deployment which relies on Amalthea: https://ci-renku-2142.dev.renku.ch/

### DIFF
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -22,6 +22,18 @@ then
 fi
 
 # install git hooks
+# we need to avoid a race condition where renku could try to install
+# the git hooks before the repo is actually cloned during the execution
+# of the entrypoint of the git sidecar
+if [[ -v GIT_CLONE_REPO && -v NOTEBOOK_DIR ]]
+then
+    while [[ ! -f ./.renku-repo-clone-complete ]]
+    do
+        echo "Waiting for repository to be cloned..."
+        sleep 1
+    done
+    rm ./.renku-repo-clone-complete
+fi
 ~/.local/bin/renku githooks install || true
 
 # run the post-init script in the root directory (i.e. coming from the image)
--- a/docker/py/entrypoint.sh
+++ b/docker/py/entrypoint.sh
@@ -1,36 +1,25 @@
 #!/bin/bash
 
-# Setup git user
-if [ -z "$(git config --global --get user.name)" ]; then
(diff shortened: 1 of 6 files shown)


docker/py/entrypoint.sh
```diff
@@ -1,37 +1,26 @@
...
-url=$(echo ${GITLAB_URL/$proto/})
-user=$(echo ${CI_REPOSITORY_URL/$proto/} | grep @ | cut -d@ -f1)
-
-git clone --depth 1 ${GITLAB_URL}/${JUPYTERHUB_USER}/renku-env.git /tmp/renku-env || true
-
-# append the contents of all the files to same files in ${HOME}
-find /tmp/renku-env -not -path '*.git*' -type f -print0 | xargs --null -I{} sh -c 'cat {} >> ${HOME}/$(basename "{}")' || true
-
 # install git hooks
 ~/.local/bin/renku githooks install || true
```
TaoSunVoyage, line 25: For R image in ci-renku-2142, though the session got started, in the logs:

> Error: . is not a renku repository.
To initialize this as a renku repository use: renku init

ableuler: Good catch. This is actually a race condition which I've introduced by cloning the repo in the entrypoint of a "normal" container rather than an init container. I did this because I wanted to use the git-https proxy to manage the git connection _during_ the clone, however, this proxy is no available during the execution of init containers. I've now removed that race condition by adding a small dummy file to the repo when the clone is done, and by waiting for that dummy file to appear inside the entrypoint of the jupyter server container.

TaoSunVoyage: thanks @ableuler, it's working now 👍

gavin-k-lee (approved):
LGTM!

PR #187, opened 2021-09-08
fix: use virtualenv manually instead of pipx for renku install

ableuler:
This PR is a reaction to a breaking change of setuptools which in turn broke our builds as pipx just pullst the latest of setuptools on a regular basis. We have therefore decided to remove pipx as it makes for non-reproducible builds.

### DIFF
--- a/docker/py/Dockerfile
+++ b/docker/py/Dockerfile
@@ -59,10 +59,14 @@ RUN conda install gxx_linux-64 && \
 # install renku-python
 ENV RENKU_DISABLE_VERSION_CHECK 1
 
-ENV PATH=$HOME/.local/bin:$PATH
+ENV PATH=$PATH:$HOME/.renku/bin
 
-RUN pipx install --pip-args="--no-cache" renku && \
-    pipx inject --pip-args="--no-cache" renku sentry-sdk
+RUN mkdir -p $HOME/.renku/bin && \
+    virtualenv $HOME/.renku/venv && \
+    source $HOME/.renku/venv/bin/activate && \
+    pip install --no-cache renku sentry-sdk && \
+    deactivate && \
+    ln -s $HOME/.renku/venv/bin/renku $HOME/.renku/bin/renku
 
 # configure git
 COPY git-config.bashrc /home/$NB_USER/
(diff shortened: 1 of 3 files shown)


docker/py/Dockerfile
```diff
@@ -61,8 +61,12 @@ ENV RENKU_DISABLE_VERSION_CHECK 1
 
 ENV PATH=$HOME/.local/bin:$PATH
 
-RUN pipx install --pip-args="--no-cache" renku && \
-    pipx inject --pip-args="--no-cache" renku sentry-sdk
+RUN mkdir -p $HOME/.local/bin && \
```
rokroskar, line 64: If the user were to install renku as a library with `pip install --user renku` - would the executable go to the same place? What happens in that case?

ableuler: This would overwrite it - bug or feature?

ableuler: ie replace the symbolic link with the actual executable.

ableuler: This makes me think - should we give the user's version precedence, even without the --user flag? we could do this by appending `$HOME/.local/bin` to the PATH rather than prepending it.

rokroskar: I think overwriting it and using the version coming from the pip install is what you would expect (then when you do `import renku` it's importing the same code as the CLI is using). Anything else would be wrong. I just wasn't sure if it would throw an error of any kind.

rokroskar: the user won't be able to install it without the --user flag... afaik

ableuler: right... at least not inside a running session. but by adding it to the project requirements.txt they can. In that case, conda will actually update the path again during the install and prepend it's location, which makes the path look like 
```
/opt/conda/bin:/opt/conda/condabin:/home/jovyan/.local/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
which gives the conda-installed version precedence. I'd have the tendency to

- create a symbolic link to location where it won't be overwritten
- add this location to the end of the path such that user-installed renku versions always win.

wdyt?

rokroskar: I think that makes sense

ableuler: done

rokroskar (approved):

PR #193, opened 2021-10-12
Fix renku not found error

TaoSunVoyage:
#187 changed where `renku` is installed.

`renku` is now in either `$HOME/.renku/bin` or `/share/bin` (batch)


ableuler:
@TaoSunVoyage this is an important fix of yours that should be merged asap. Could you please follow up on it and make sure it goes in? 🙏

TaoSunVoyage:
@ableuler sure, I am checking why the checks failed

ableuler:
Ah, I just realise that the tests are failing here because #197 is not yet merged.

PR #197, opened 2021-10-25
Fix image build and bump base image version

ableuler:
This bumps the notebooks base image from `lab-3.0.16` to `lab-3.2.1`, so a good look from the @SwissDataScienceCenter/poc squad to check if that breaks something would be appreciated.

### DIFF
--- a/docker/batch/Dockerfile
+++ b/docker/batch/Dockerfile
@@ -1,5 +1,5 @@
 # define build arguments
-ARG RENKU_BASE=renku/renkulab-py:3.9-0.10.1
+ARG RENKU_BASE=renku/renkulab-py:latest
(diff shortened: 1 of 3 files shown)


ableuler:
@TaoSunVoyage any chance you could give it a try with the fix that you and @gavin-k-lee have discussed. We should try and get this fix to the users asap.

TaoSunVoyage:
The simple copy doesn't work as the python executable in the venv is a symbolic link to the conda python. So, I reinstalled it in batch image instead.

ableuler (commented):
LGTM, @TaoSunVoyage  feel free to approve and merge.

TaoSunVoyage (approved):

PR #202, opened 2021-11-15
fix: use renku cli from PATH

ableuler:
After getting rid of pipx, the renku cli executable is in a different location. Therefore, it is not found anymore during the execution of the entrypoint. Since we already make sure that the PATH contains the location of the renku cli, we should omit the exact path to the renku cli executable and instead just use whatever is found in the PATH.


ableuler:
closing as this is already handled by #193

PR #212, opened 2021-12-15
feat: support multi-platform build using buildx

vigsterkr:
Currently only `linux/amd64` docker images are being built, which causes the arm users to run the whole image in an emulator... this adds support for multi-platform builds.


vigsterkr:
@SwissDataScienceCenter/yat i'm wondering if it makes sense to build all the platforms for the acceptance test as well. coz that'll adds some significant amount of build-time because of `qemu`

rokroskar:
It looks like a `docker push` that is separate from the `buildx` command pushes just a single arch image, not the multi-arch. Maybe `docker buildx build --push` has to be used always?

rokroskar:
Actually thinking more about this, we don't gain a lot by making the base images multi-platform because no one uses the base images directly. If anything, the gitlab ci jobs in the Renku projects themselves should make multi platform images so they can be used for local sessions. WDYT @olevski?

olevski:
@rokroskar I agree, I think the majority of users will not need these. But if the CI pipelines in projects build multi platform images then wouldn't they need to start with these multi platform images? I.e. the base images would still be needed.

So if want to support multi platform images fully then we should do this PR and then also modify the project templates. Or mention in the docs how one can do this with gitlab ci.

rokroskar:
I also thought so but it doesn't seem to be the case - basically you don't need everything in the dependency chain to be multi-arch.

rokroskar:
haha well of course I'm talking nonsense. I think we should finish this PR... the base images for the jupyter stack are multi-arch so at least for our `renku/renkulab-py` image the multi-arch build makes sense. Rocker images don't seem to be multi-arch so for anything R-based it doesn't make sense to do multi-arch for now.

rokroskar:
btw the reason I went down this rabbit hole again is because it turns out streamlit doesn't work on m1 running via qemu... so I had to build a native image to get it to work locally. 1.5 hours later I got it working... would have saved some time if I had a native base image 😅 Initially I thought the only issue would be a performance penalty but it turns out that some code just won't run.

docker/py/Dockerfile
```diff
@@ -1,5 +1,37 @@
...
+    # npm cache clean --force && \
+    rm -rf "/home/${NB_USER}/.cache"
+
+# jupyter sets channel priority to strict which often causes very long error messages
+RUN mamba config --system --set channel_priority flexible && \
+    mamba clean --all -f -y
+
+COPY renku-requirements/requirements.txt /tmp/renku-requirements.txt
+
+RUN mkdir -p "$HOME/.renku/bin" && \
```
olevski, line 28: @rokroskar why dont we move the linux package installation in the base image here? I know we discussed this briefly but how did you decide what goes in the first stage?

olevski (changes requested):
One question about the multi-stage builds. I think we can push more things into the 1st stage?

olevski (changes requested):
One question about the multi-stage builds. I think we can push more things into the 1st stage?

EDIT: I answered my own question. Things are just copied into the last stage from the build stage. So there are no other steps that are as easy to move into the build stage other than the python virtual env.

docker/py/Dockerfile
olevski: Nvm I remembered our discussion. You just want to skip the python env setup with the build stage. But all other things are installed in the final stage.

rokroskar: Only the python install is done here - the apt is unfortunately needed to install the compiler (there is no arm wheel for psutil afaict)

rokroskar: basically we can copy the entire content of what gets installed for the python dependencies because it all goes into the conda directory. The rest is not easy to transfer between stages. I thought about streamlining the big apt install (we probably don't _need_ all of that) but I left it for another day.

olevski (approved):

rokroskar:
Thanks for reviewing @olevski and thanks for kicking this off all those months ago @vigsterkr!! 🙌

Issue #160, opened 2021-05-21
R 4.1.0 is not compatible with RStudio 1.3.959 (plots)


gavin-k-lee:
![image](https://user-images.githubusercontent.com/22785147/119164279-89683e80-ba5c-11eb-9114-b3313e2ffbb5.png)

Plotting does not work.

R 4.0.5 plotting does work with RStudio 1.3.959 though.

Related to #154 (known issue).

gavin-k-lee:
This also blocks adding Bioc 3.13 (released this weekend) since that is built against R 4.1.

PR #218, opened 2022-01-25
chore: upgrade R and RStudio versions

gavin-k-lee:
We have been out-of-date with R/RStudio for a while now (#154 & #160) due to external dependencies. Since these external dependencies have been resolved, we can build newer versions based on Jupyter proxies.


olevski:
@gavin-k-lee I think all the pieces (i.e. dependent libraries) that needed new releases for this to work are now done. I did some tests in a different branch. Are you ok for me to transfer these changes here so that we can have images with newer R and Rstudio versions?

olevski:
Some related issues we can close with this: #154 and #160.

gavin-k-lee:
@olevski yes please! transfer away. thank you!!

docker/r/Dockerfile
```diff
@@ -31,8 +31,9 @@ RUN echo "PATH=${PATH}" >> /usr/local/lib/R/etc/Renviron && \
 ENV LD_LIBRARY_PATH /usr/local/lib/R/lib
 
 # pin the version of RStudio
-ENV RSTUDIO_VERSION 1.3.959
-RUN /rocker_scripts/install_rstudio.sh
+# ENV RSTUDIO_VERSION 2021.09.2+382
+# RUN /rocker_scripts/install_rstudio.sh
+# This version of rocker should already have RStudio installed!
```
rokroskar, line 36: it's probably a good idea to pin the version of RStudio to avoid surprises anyway, even if it's just to the one that this rocker image is shipping with atm.

gavin-k-lee: yep. btw after 1.4.XYZ they moved to calendar versioning: https://www.rstudio.com/products/rstudio/release-notes/

olevski:
Also this is a good guide on the insanity that is R session environment variables. If I change only the `Renviron` file the path is not updated. I could only get it to work when I change `Renviron.site` - and that is what is actually recommended.

See: https://rviews.rstudio.com/2017/04/19/r-for-enterprise-understanding-r-s-startup/

rokroskar:
@olevski yes the RStudio environment setup is mind-boggling. It basically ignores any reasonable convention that you might expect 😭 🤦

rokroskar:
Though I'm a bit confused why adding the paths to `Renviron.site` is needed - it _used_ to work as is

gavin-k-lee:
btw I think the bioc builds might fail since the R and RBioc versions aren't corresponding exactly. 
Bioc 3.11 ~ R 4.0.0
Bioc 3.12 ~ R 4.0.3
Bioc 3.13 ~ R 4.1.0
Bioc 3.14 ~ R 4.1.1

olevski:
@gavin-k-lee the bioc builds were failing because of mistakes in the RSTUDIO_VERSION environment variable not being properly defined in the dockerfile. They should all pass now.

olevski:
Ok this is ready to go now.

Here is a test project with bioconductor 13: https://dev.renku.ch/projects/tasko.olevski/test-image-upgrade-r-1

olevski:
Here is a project with just R 4.1.0: https://dev.renku.ch/projects/tasko.olevski/test-r-project-1

olevski:
Both of the test projects work!

gavin-k-lee:
nice! I just tried both projects and when opening in an external tab it worked great and I tested the plotting (which was the most prominent problem before). In the iframe RStudio took a long while to load and I had to fresh the session page a couple of times - not sure if you experience that all? Maybe it is a resource issue?

olevski:
@gavin-k-lee I have seen the loading issues in the iframe you mention on other versions (in production) too. I am pretty sure it is related to resource usage. I never saw it in test but then I made sure to request 1cpu with the sessions - not sure if that helped really. I think the cpu should not be limited in k8s.

Either way I think this is a problem that shows up sometimes and is not a result of the new images.

olevski:
On another note... 

I def lost some sanity just trying to get these images to build btw. I did notice that building the dockerfiles worked differently on my mac locally and in github CI. But then after I switched to my personal latptop (which runs linux) I could finally get the same outcomes I see in the CI workflows locally and eventually fix them. Just a PSA if some else is having such weird issues.

gavin-k-lee:
yeah it seems like you really need to build them with linux to have them equivalent, which is annoying.

@SwissDataScienceCenter/renku-ui-maintainers would we be able to get some help to understand what's going on with the iframes? I tried the highest cpu and ram on dev and the iframe didn't load (but the new tab worked as expected)

gavin-k-lee:
I think we can also retire some of the 4.0.X images and bioc images. We could go with:
R 4.1.0, 4.1.1, 4.1.2
Bioc 3.13, 3.14.

R 4.0.5 came out in March 2021 and Bioc 3.12 in October 2020.

Thoughts?

ciyer:
I just tried both the test-image-upgrade and the test-r-project above, and I did **not** see anything unusual in the iframe experience. Both projects took 40-45s to start, but I had the spinner/bouncer and then they transitioned seamlessly to RStudio once the pods were up. It seems that the problem is intermittent.

If you could note the time (UTC) when you next have issues, we can see if there is anything in the logs that sheds light on what is going on.

olevski:
@rokroskar I tested the latest version of the R images with the `devel` images for R and bioconductor. Things work. So this is ready to go!

If you or anyone else is interested here are the projects:
- https://dev.renku.ch/projects/tasko.olevski/test-image-upgrade-r-1 (plain R)
- https://dev.renku.ch/projects/tasko.olevski/test-r-project-1 (bioconductor)

rokroskar (approved):
this is awesome, thanks @gavin-k-lee and @olevski !!!

PR #222, opened 2022-01-31
feat: add matlab enabled Docker image

gavin-k-lee:
After much demand from the community and some prototypes (see [here](https://renkulab.io/gitlab/learn-renku/working-with-matlab)) we should add a Matlab-enabled Renku docker image. This simplifies down the Docker images and we can then make templates from this image.


docker/matlab/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+    zlib1g \
+    && apt-get clean \
+    && apt-get -y autoremove \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install jupyter-matlab-proxy dependencies
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install --yes \
+        xvfb \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
rokroskar, line 114: why is this separate from the other `apt-get` just above?

rokroskar (changes requested):
Thanks! Just a minor thing about apt-get ordering

docker/matlab/Dockerfile
gavin-k-lee: no real reason - it was taken from the dockerfile provided by matlab

gavin-k-lee:
Test repo here: I'm having trouble getting it to work as I did before... the logs show something like this:
```
INFO:MATLABProxyApp:Waiting for MATLAB to exit...
INFO:MATLABProxyApp:MATLAB has exited with errorcode: -9
ERROR:MATLABProxyApp:MATLAB returned an unexpected error. For more details, see the log below.

[W 23:18:07.147 NotebookApp] 404 POST /sessions/lee-2egavi-test-2dmatlab-2ddocker-6e65d600/matlab/messageservice/json/secure (127.0.0.1) 70.430000ms referer=https://renkulab.io/sessions/lee-2egavi-test-2dmatlab-2ddocker-6e65d600/matlab/remote/proxy/1.6/iframeProxyRelease.html
```
Repo: https://renkulab.io/projects/lee.gavin.k/test-matlab-docker/ (log into MATLAB with your ETH-provided license, however this might be the issue according to https://github.com/mathworks/jupyter-matlab-proxy/blob/main/MATLAB-Licensing-Info.md).

I tried a previous matlab build (it is a slightly different matlab version) here: https://renkulab.io/gitlab/learn-renku/working-with-matlab and it works fine. It could also be the security settings that changed with regards to iframes recently?

gavin-k-lee:
I just tried with the VNC (without the jupyter-matlab-proxy) and it seems to play ball - I think the new security settings + the ways the ports are set up in the jupyter-matlab-proxy make it unhappy.

I will reformulate this.

rokroskar:
What do you mean by "new security settings"?

rokroskar:
I just tried the test-matlab-docker and it seems to be ok? Or maybe I'm not looking at the right things?

gavin-k-lee:
New security settings: the ones renku-ui put in to disallow some particular js/browser-based features like interactive plots.

Yes I believe I've fixed it, so the latest test-matlab-docker should work.

rokroskar:
yeah but those settings just disallow rendering of JS / iframes in notebook previews. It shouldn't have any effect on the session window afaik, if it does it's a bug.

gavin-k-lee:
Yeah it probably was a bug on matlab's side. Anyway I think this one is alright now - you would access it through the vnc directly.

rokroskar:
So I just tested the latest image and it definitely seems to work - but in the previous versions you had matlab opening up directly - is that no longer needed? My thinking was just if you could open it directly via a proxy then you actually don't need all the vnc stuff. But if this is how it's intended/useful it's also fine.

rokroskar:
also - this PR definitely deserves a description! 😆

gavin-k-lee:
Yeah, so in the previous version it was using a slightly older matlab docker base, and the vnc-proxy worked. However when I updated it it started bugging and shut down after I entered my license details. So I opted to use our renku-vnc and have users open it from the vnc.

rokroskar:
ok all good - I'll approve then. But please do add a description for posterity ;)

rokroskar (approved):
Thanks @gavin-k-lee I imagine this will make lots of people happy!

gavin-k-lee:
@rokroskar sorry for the commit spam - I was trying to get the desktop icon to play nicely but it seems like it would be another rabbit hole to go down if I was to find out the root problem (I think it's down to environment variables). Basically now when you click on the desktop icon you get taken to the /opt/VNC folder but even when I passed in the starting directory, matlab would complain. I don't think it's a huge issue and we can instruct them to navigate to their project git repo in the container.

rokroskar:
Sounds good - we can try to fix that in a future iteration, I think it's already useful as-is

rokroskar (approved):

PR #225, opened 2022-02-01
chore: bump renku version

rokroskar:
version bump - normally this isn't necessary but since we're releasing a lot of things all at once might as well make it consistent.


gavin-k-lee (approved):
LGTM!

PR #230, opened 2022-02-17
feat: use jupyterlab from base image

noeleont:
Since jupyterlab is already included in the base base image it should be
possible to use this version instead of installing an older version afterwards.


rokroskar:
Hi @noeleont thanks for the PR - the pinned version here is just to `3.0.x` and that's to avoid any surprises with incompatible updates. Usually it's a problem with the plugins. If we verify that the plugins work with an updated version, then we can bump the version range.

rokroskar:
Hi @noeleont thanks for the PR - the pinned version here is just to `3.0.x` and that's to avoid any surprises with incompatible updates. Usually it's a problem with the plugins. If we verify that the plugins work with an updated version, then we can bump the version range.

rokroskar:
@noeleont could you verify that the `jupyterlab-system-monitor` and `jupyterlab-git` extensions still work in this case? If so, then I think we can go ahead and merge this.

rokroskar:
@noeleont I will close this  PR because I cherry-picked your commit into another branch so that the CI will run. Thanks for pointing this out, I will test the compatibility of the various libraries in the other PR.

PR #232, opened 2022-03-17
fix: add renku cli to PATH in rstudio

olevski:
Fixes renku not being present in the PATH of rstudio images.

### DIFF
--- a/.github/workflows/build_and_push_to_docker.yml
+++ b/.github/workflows/build_and_push_to_docker.yml
@@ -24,8 +24,8 @@ jobs:
     - name: Docker Login
       uses: Azure/docker-login@v1
       with:
-        username: ${{ secrets.DOCKER_USERNAME }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
+        username: ${{ secrets.RENKU_DOCKER_USERNAME }}
+        password: ${{ secrets.RENKU_DOCKER_PASSWORD }}
 
     - uses: actions/checkout@v2
     - name: Build renku project python-based docker images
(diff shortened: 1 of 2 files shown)


olevski:
This project: https://renkulab.io/projects/tasko.olevski/test-r-project-1 uses the image built in this repo. And renku cli works in both rstudio and jupyterlab.

rokroskar:
if I'm reading the R Dockerfile correctly, we never actually install renku in there. Should we copy it from the python base image like we do the other python packages [here](https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/docker/r/Dockerfile#L75-L80)? I guess all that is needed is something like 

```
COPY --from renku_base /home/jovyan/.renku /home/rstudio/.renku
```

olevski:
> if I'm reading the R Dockerfile correctly, we never actually install renku in there. Should we copy it from the python base image like we do the other python packages [here](https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/docker/r/Dockerfile#L75-L80)? I guess all that is needed is something like
> 
> ```
> COPY --from renku_base /home/jovyan/.renku /home/rstudio/.renku
> ```

Rok what you describe does happen in the Dockerfile here: https://github.com/SwissDataScienceCenter/renkulab-docker/pull/232/files#diff-0905cf205d369bd74981475a9ea32b552da07810e4010c9936058a9837a6885eR79.

Maybe I misunderstand what you mean though.

gavin-k-lee:
I verified that the cli is available in that image :+1:

rokroskar:
@olevski you're totally right I missed it because of the line break 🤦 !!

olevski:
@rokroskar @gavin-k-lee if you are done testing/reviewing can you approve please?

rokroskar (approved):

PR #234, opened 2022-03-21
chore: prep 0.11.1 release

olevski:
Most notably adds the renku cli in the path for R based images


rokroskar (approved):
Thanks @olevski !

PR #244, opened 2022-06-14
fix: R image builds

olevski:
The reason for this is that the rocker/verse:devel image has switched to ubuntu 22.04 (jammy) but the rstudio .deb installer is only available for focal and older versions. The rstudio installation script builds the url from the ubuntu version and it just fails.

The rstudio version that is available for ubuntu 22.04 is `2022.02.3-492`. So I have switched the ci pipeline to use that for the devel image.

I will test if this work with renku.

### DIFF
--- a/.github/workflows/build_and_push_to_docker.yml
+++ b/.github/workflows/build_and_push_to_docker.yml
@@ -353,12 +353,17 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        RVERSION:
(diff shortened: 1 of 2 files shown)


rokroskar:
I checked the devel image - it works well. Maybe we can bump the RStudio version for the other builds too in a follow-up.

rokroskar (approved):
Thanks @olevski!!!

PR #247, opened 2022-06-15
chore: add linting for Dockerfiles with hadolint

olevski:
It is a good idea to start doing this and enforcing it.


docker/vnc/Dockerfile
```diff
@@ -54,20 +53,16 @@ RUN curl -sSfL https://sourceforge.net/projects/tigervnc/files/stable/${tigervnc
...
 COPY .git_icon.png /home/jovyan/Desktop/
-RUN chmod +x /home/jovyan/Desktop/gitk.desktop
-
-#################################################################
-# Add wallpaper and force the default to point to the renku one
-
-RUN mkdir -p /usr/share/backgrounds/renku/
 COPY renku_background_dots.png /usr/share/backgrounds/renku/
-RUN ln -s -f /usr/share/backgrounds/renku/renku_background_dots.png /usr/share/backgrounds/xfce/xfce-stripes.png
+RUN chmod +x /home/jovyan/Desktop/gitk.desktop && \
```
olevski, line 62: No real changes - just reorganized commands so that we merge them into one large RUN command instead of having separate ones.

olevski:
@rokroskar the builds and linting is passing. So we can merge this. The only changes I made are to pass the linting errors.

The linting warnings I am ignoring are:
- DL3006 | Warning | Always tag the version of an image explicitly. - we actually do this but hadolint does not recognize it because we use an `ARG` and then we set that arg to a specific version at build time
- DL3008 | Warning | Pin versions in apt-get install. - I think this is nice but would potentially break builds as different base images are used, we could also make an issue to not ignore this issue and pin everything
- DL3013 | Warning | Pin versions in pip. - similar to above
- SC1091 - gets triggered on `RUN source ...` commands in the Dockerfile when hadolint cannot find the file that is sourced locally - this can definitely be ignored IMO, it does not even appear in the official list on the hadolint repo readme

rokroskar (approved):
🚀

PR #248, opened 2022-06-16
chore: add tensorflow 2.9 image

olevski:
Adds a cuda image with tensorflow 2.9.

As per a user request https://renku.discourse.group/t/tensorflow-2-9-0/642


rokroskar (approved):
Lets merge after the user confirms that it works

PR #250, opened 2022-07-01
chore: update git LFS to 3.2.0 for python images

olevski:
The really old git lfs version is actually not working on the BIT deployment. But updating to a newer version works without a problem.

Even if this was not the case using a newer version of git LFS is a good idea.


rokroskar:
I believe it was pinned because of issues with renku-python - can you maybe run renku-python unit tests with this version?

olevski:
@rokroskar [here](https://github.com/SwissDataScienceCenter/renku-python/pull/2985) is a dummy PR to trigger the tests. Not sure if the changes I did are enough to make sure the tests run wiht the latest git lfs version. I asked Ralf to confirm.

rokroskar (approved):

PR #251, opened 2022-07-22
chore: library upgrade

rokroskar:
upgrades the libraries pinned in the base image.


rokroskar:
@olevski @gavin-k-lee a bunch of things are changed here in these images so it would be great if you could give them a once-over. They seem fine to me. Note also that since the default python version in the latest `jupyter/base-notebook`images is 3.10, we now inherit this python version as well. We'll need to update the templates and probably announce it to the users in the release notes.

docker/py/Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG BASE_IMAGE=jupyter/base-notebook:lab-3.2.1
-ARG RENKU_VERSION=1.4.0
+ARG BASE_IMAGE=jupyter/base-notebook:lab-3.4.4
+ARG RENKU_VERSION=1.5.0
```
lorenzo-cavazzi, line 2:
```suggestion
ARG RENKU_VERSION=1.6.0
```

Panaetius:
According to https://docs.python.org/3/whatsnew/3.9.html#you-should-check-for-deprecationwarning-in-your-code 3.9 is the last version to contain 2.7 backwards compatibility layer and https://docs.python.org/3/whatsnew/3.10.html#removed did in fact remove some module methods and quite a lot of C API functions/types (at the end of the page). So 3.10 probably has a bit higher chance of breaking things than was the case with 3.9

We do NOT have tests for 3.10 in renku-python, so upgrading to 3.10 would essentially be flying blind, plus the higher risk of breaking user code does mean it'd make sense to upgrade to a version prior to https://github.com/jupyter/docker-stacks/blame/0be9e335026b21a452c888481ab6341dcdd183a5/base-notebook/Dockerfile#L88

rokroskar:
Something is wrong with the `cuda-tf` builds - also failing on master. Will make an issue to fix it.

docker/py/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash && \
     apt-get update && apt-get install -yq --no-install-recommends \
     git \
-    git-lfs=3.2.0 \
+    git-lfs=3.1.4 \
```
olevski, line 17: Why not leave this at 3.2.0?

olevski (dismissed):

olevski (approved):

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -62,7 +62,7 @@ jobs:
 
         # This ensures the same image tags as before are built, in addition to new ones
         if [[ "${{ matrix.BASE_IMAGE_TAG }}" == lab-* ]]; then
-          export LABEL_PREFIX=3.9
+          export LABEL_PREFIX=3.10
```
olevski, line 65: @rokroskar I thought none of the images are using python 3.10. And from Ralf's comment it seems we are not ready for 3.10 yet.

So this should then probably be rolled back to 3.9?

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -356,7 +356,7 @@ jobs:
 
         export DOCKER_TAG="${{ matrix.JULIAVERSIONS }}-$LABEL"
         docker build docker/julia \
-          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.9-$LABEL" \
+          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.10-$LABEL" \
```
olevski, line 359:
```suggestion
          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.9-$LABEL" \
```

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -408,11 +408,11 @@ jobs:
 
         export DOCKER_TAG="${{ matrix.RVERSION }}-$LABEL"
         docker build docker/r \
-          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
+          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.10-$LABEL" \
```
olevski, line 411:
```suggestion
          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
```

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -462,7 +462,7 @@ jobs:
 
         export DOCKER_TAG="${{ matrix.RELEASE }}-$LABEL"
         docker build docker/r \
-          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
+          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.10-$LABEL" \
```
olevski, line 465:
```suggestion
          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
```

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -320,8 +320,8 @@ jobs:
         fi
 
         docker build docker/batch \
-          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
-          --build-arg BASE_IMAGE="python:3.9-slim-buster" \
+          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.10-$LABEL" \
```
olevski, line 323:
```suggestion
          --build-arg RENKU_BASE="$DOCKER_NAME-py:3.9-$LABEL" \
```

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -320,8 +320,8 @@ jobs:
...
+          --build-arg BASE_IMAGE="python:3.10-slim-buster" \
```
olevski, line 324:
```suggestion
          --build-arg BASE_IMAGE="python:3.9-slim-buster" \
```

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -114,7 +114,7 @@ jobs:
         fi
 
         docker build docker/$EXTENSION \
-          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.9-$LABEL" \
+          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.10-$LABEL" \
```
olevski, line 117:
```suggestion
          --build-arg BASE_IMAGE="$DOCKER_NAME-py:3.9-$LABEL" \
```

.github/workflows/build_and_push_to_docker.yml
olevski:
```suggestion
          export LABEL_PREFIX=3.9
```

olevski (changes requested):

PR #254, opened 2022-07-27
fix: label for python in docker images

olevski:


rokroskar (approved):
Thanks for cleaning up my mess :)

PR #258, opened 2022-08-09
fix: qgis image build

olevski:
It seems that the gpg key was expired.

The qgis [docs](https://www.qgis.org/en/site/forusers/alldownloads.html#quickstart) show a new key that should be used.


lorenzo-cavazzi (approved):
:rocket:

PR #265, opened 2022-09-18
chore(ci): common tests library and rstudio tests

olevski:
Since I use the same cypress specs here and in the amalthea repo. I created a separate npm package with these.

Then I do not have to have copies of the same cypress tests in two different places.

In addition to this I added a spec to do tests for rstudio.

closes https://github.com/SwissDataScienceCenter/amalthea/issues/192


rokroskar (approved):
🤖

PR #267, opened 2022-09-26
feat: add python 3.10 base image

olevski:


docker/py/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     musl-dev \
     nano \
     netcat \
-    python-dev \
```
rokroskar, line 31: is this even necessary?

olevski: I tried taking it out and some of the lower python version images were failing I think. I can double check. I re-wrote that commit fully unfortunately. I should have let it stick around.

olevski: @rokroskar you are right python-dev is not really needed.

rokroskar (approved):

PR #271, opened 2022-10-25
fix: broken nbconvert package

olevski:
This was causing a lot of warnings in the console and making certain features of sessions buggy/unstable.

For example converting a notebook into anything would fail.

But also the notebooks and the jupyter server uses nbconver (for some reason) while it runs.

Example of failed client calls with our latest version:
image

Also failures such as this show up in the browser console:
image

With the update there are no such failed requests and the console has only warnings.

Here is a project that has the image from this PR pinned in its latest commit. In the first commit the older image that has issues is used.

There is an option to enable cypress to fail when there are any failures in the console. That would have caught this problem. I will enable it in a separate PR.

### DIFF
--- a/docker/py/requirements.txt
+++ b/docker/py/requirements.txt
@@ -6,3 +6,4 @@ powerline-shell~=0.7.0
 requests>=2.20.0
 virtualenv>=20.7.2
 mistune>=2.0.1 # not directly required, pinned by Snyk to avoid a vulnerability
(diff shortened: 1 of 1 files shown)


olevski:
p.c. the image from this PR is registry.dev.renku.ch/tasko.olevski/renku-images/renkulab-docker:0.0.1

Also here is the related issue on jupyter: https://github.com/jupyter/nbconvert/issues/1742

rokroskar (approved):

PR #277, opened 2022-11-28
chore: amend build to use dependabot for renku updates

rokroskar:
This PR sets up the renkulab-docker repo to automatically update the renku CLI version with dependabot. 

The main change is that a requirements file is now used to set the renku-python version. This way it can automatically be bumped by dependabot. In addition, any images that also need to install renku-python simply grab it from the base renku image.


docker/py/Dockerfile
```diff
@@ -42,7 +41,7 @@ RUN groupadd -g 1000 jovyan
 USER $NB_USER
 
 # install jupyterlab, papermill, git extension and renku-jupyterlab-ts
-COPY requirements.txt /tmp/requirements.txt
+COPY docker/py/requirements.txt /tmp/requirements.txt
```
olevski, line 44: @rokroskar why did you change the context for the builds above?

I.e. we had `docker build <path to docker image folder>` but you switched it to `docker build . -f <full_path_to_dockerfile>`.

I think it makes more sense for the build context to be the location of each specific `Dockerfile` and not the root of the repository.

Also with this change now we have some images whose build contexts are in the same folder as the Dockerfile and some where the build context is `.` (i.e. the root of the repo).

I think it is best to keep this consistent and to keep things as they were. The build context for each image is the folder where the Dockerfile lives and not the root of the repository.

.github/workflows/build_and_push_to_docker.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           export LABEL_PREFIX=${{ matrix.BASE_IMAGE_TAG }}
         fi
 
-        docker build docker/py --tag $DOCKER_NAME-py:$LABEL_PREFIX-$LABEL --build-arg BASE_IMAGE="jupyter/base-notebook:${{ matrix.BASE_IMAGE_TAG }}"
+        docker build . -f docker/py/Dockerfile --tag $DOCKER_NAME-py:$LABEL_PREFIX-$LABEL --build-arg BASE_IMAGE="jupyter/base-notebook:${{ matrix.BASE_IMAGE_TAG }}"
```
olevski, line 71: See my other comment. I think this should remain as is. The build context should be the location where the Dockerfile for each image is. With this change we make the context be the root of the repository and then we have to include/change paths to every file referenced in the Dockerfile relative to the root of the repository.

olevski (changes requested):

docker/py/Dockerfile
rokroskar: yeah this is a bit dirty, I agree, and I don't really like it. But I want to have one location for the renku dependency for any of the images that need it.

rokroskar: The alternative is to have the requirements file in multiple places. Not a horrible solution but not a great one.

rokroskar: @olevski I went back to the previous structure. Please have a look 🙏

olevski (approved):

PR #290, opened 2023-01-10
fix(build): fix makefile and add some documentation

seanrmurphy:


docker/cuda/local_build.sh
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env bash
```
rokroskar, line 1: why use this when there is a `Makefile`? The cuda libraries should be set to some sensible default in the `Makefile` and overridable via environment variables, exactly the same way as here.

rokroskar (changes requested):
Thanks @seanrmurphy! I think it would be nice to get the cuda build through the Makefile in the same way as your script now does.

docker/cuda/local_build.sh
seanrmurphy: Sure - I only looked at the makefile afterwards; in fairness, the instructions say to use docker in the directory so that's where I went first..

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -1,6 +1,12 @@
 name: Renku Docker Image CI
 
-on: [push]
+on: 
+  #  pull_request:
+  #    branches: [master]
+  #    paths-ignore: ['Makefile', 'docker/**']
+  push:
+      branches: ['*']
```
rokroskar, line 8: is this needed?

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -182,27 +151,28 @@ jobs:
...
+        export DOCKER_PREFIX="renku/renkulab"    
+        export DOCKER_LABEL="${{ matrix.CUDA_VERSION }}-${LABEL}"
+        export RENKU_BASE="renku/renkulab-py:python-${{ matrix.PYTHON_VERSION }}-$LABEL" 
+        export CUDA_VERSION="${{ matrix.CUDA_VERSION }}" 
+        export EXTRA_LIBRARIES="${{ matrix.EXTRA_LIBRARIES }}" 
+        export CUDA_CUDART_PACKAGE="${{ matrix.CUDA_CUDART_PACKAGE }}" 
+        export CUDA_COMPAT_PACKAGE="${{ matrix.CUDA_COMPAT_PACKAGE }}" 
+        export LIBCUDNN_PACKAGE="${{ matrix.LIBCUDNN_PACKAGE }}" 
+        make cuda
+        echo "IMAGE_NAME=$DOCKER_NAME-cuda:$DOCKER_LABEL" >> $GITHUB_OUTPUT
```
rokroskar, line 175:
```suggestion
        echo "IMAGE_NAME=$DOCKER_PREFIX-cuda:$DOCKER_LABEL" >> $GITHUB_OUTPUT
```

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -68,9 +73,15 @@ jobs:
...
         fi
 
-        docker build docker/py --tag $DOCKER_NAME-py:$LABEL_PREFIX-$LABEL --build-arg BASE_IMAGE="jupyter/base-notebook:${{ matrix.BASE_IMAGE_TAG }}"
-        echo "IMAGE_NAME=$DOCKER_NAME-py:$LABEL_PREFIX-$LABEL" >> $GITHUB_OUTPUT
-        
+        # needed by the makefile
+        export DOCKER_PREFIX="renku/renkulab"    
+        export DOCKER_LABEL="${LABEL_PREFIX}-${LABEL}"
+        # this is required by the makefile
+        export GIT_MASTER_HEAD_SHA=$(git rev-parse --short=7 --verify HEAD)
```
rokroskar, line 80: can't this be shared by all steps somehow?

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -277,6 +245,42 @@ jobs:
...
+          export LABEL=$(echo ${{ github.sha }} | cut -c 1-7)
+        fi
+
+        # needed by the makefile
+        export DOCKER_PREFIX="renku/renkulab"    
+        export DOCKER_LABEL="${LABEL}"
+        # this is required by the makefile
+        export GIT_MASTER_HEAD_SHA=$(git rev-parse --short=7 --verify HEAD)
+        make vnc 
+        docker push $DOCKER_NAME-$EXTENSION:$LABEL
```
rokroskar, line 280:
```suggestion
        docker push $DOCKER_PREFIX-$EXTENSION:$LABEL
```

.github/workflows/build-and-push-to-docker.yml
rokroskar: Oh I see what's going on - `DOCKER_NAME` is defined above but `DOCKER_PREFIX` is used in the `Makefile` - why not just one or the other?

.github/workflows/build-and-push-to-docker.yml
seanrmurphy: This has changed a little; I figured that it could be possible to do this but I think I've already spent enough time on this for now.

rokroskar: nonsense! 1 week minimum for PRs with with CI and docker builds

rokroskar (changes requested):
This is great, thanks @seanrmurphy ! Just found a few typos then it's gtg

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -15,16 +38,14 @@ jobs:
         - docker/py/Dockerfile
         - docker/r/Dockerfile
         - docker/cuda/Dockerfile
-        - docker/cuda-tf/Dockerfile
-        - docker/generic/Dockerfile
```
rokroskar, line 19: did you mean to remove this one?

seanrmurphy: Yes - `generic` is not referenced in the `Makefile` ; it is referenced in the github action but just for linting afaict - it does not look like an image that is actually built.

rokroskar (approved):

PR #295, opened 2023-01-19
chore: install git-lfs from tarball

### LINKED ISSUE #294: git-lfs should be installable also on arm
Currently installing `git-lfs=3.2.0` fails on arm which makes it hard to do multi-platform builds of our base images. We should find a solution to this, potentially installing from source?

### LINKED ISSUE #300: bleach==5.0.0 makes jupyter server proxy fail
With `bleach==5.0.0` the jupyter server proxy fails to load because of a bad dependency string. 

The error is

    pkg_resources.extern.packaging.requirements.InvalidRequirement: Expected closing RIGHT_PARENTHESIS
        tinycss2 (>=1.1.0<1.2) ; extra == 'css'

and seems to be a result of [this commit](https://github.com/mozilla/bleach/commit/cb6dca73d0d15f8c26acb28e7cc53482526bbe1b#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7). The manifestation of this are failing R image builds (because they rely on the jupyter server proxy for rstudio to display properly).

rokroskar:
This PR installs git-lfs from a tarball to avoid issues with missing arm debian packages for git lfs 3.x. 

closes #294 
closes #300


olevski (dismissed):

rokroskar:
Don't ask me why, but somehow yesterday a broken version of `bleach` started to get installed which broke the jupyter-server-proxy which broke the rstudio images. I've excluded the bad `bleach` version now explicitly and it seems to work...

PR #309, opened 2023-02-27
chore: use mamba everywhere

### LINKED ISSUE #282: replace conda with mamba
The title says it all.

Conda has many issues, mamba is a drop-in replacement that solves them and improves performance.

rokroskar:
Replaces `conda install` with `mamba install` everywhere. 

closes #282

### DIFF
--- a/.github/workflows/build-and-push-to-docker.yml
+++ b/.github/workflows/build-and-push-to-docker.yml
@@ -59,9 +59,9 @@ jobs:
       fail-fast: true
       matrix:
         BASE_IMAGE_TAG:
-          - lab-3.4.0
-          - python-3.10.6
-          - python-3.9.12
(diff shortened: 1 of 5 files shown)


rokroskar:
Something is wrong with the libraries required for the VNC image - I think we should fix that in a separate issue as it doesn't affect the most commonly used images.

Panaetius:
> Something is wrong with the libraries required for the VNC image - I think we should fix that in a separate issue as it doesn't affect the most commonly used images.

The VNC image is an odd one out anyways, we have it as one of our dockerfiles, but the relevant template for it is a contributed templated, so it's a bit of a mix of supported by us and not officially supported by us.

Panaetius (approved):
:snake:

PR #310, opened 2023-02-27
chore: bump jupyter base image

rokroskar:
Bumps base image. Makes the change to python 3.10.


rokroskar:
closing - will do as part of #309

PR #313, opened 2023-03-01
fix: image tagging consistency

rokroskar:
Following the Makefile refactoring there were some issues with image tags - this PR sets up a consistent use of `DOCKER_LABEL` and other environment variables.


seanrmurphy (approved):
I did some testing by setting up different configurations of env vars and seeing if the images were tagged correctly. I did this for the python, R and julia images - everything worked as expected.

rokroskar:
Thanks for the careful review @seanrmurphy !

PR #320, opened 2023-03-17
chore: override notebook command

rokroskar:
This PR injects a function into the entrypoint to override the jupyter command so that it can be forward compatible with newer images that no longer start the full server with `jupyter notebook`, making `/lab` yield a 404.


olevski (approved):
Thank you Rok!

PR #324, opened 2023-03-26
chore: ensure correct PATH on ssh

rokroskar:
`PATH` was set in some places via environment, which means it was not set on SSH. This PR fixes the problem. 

To reproduce the original issue, run a session with the `renku/renkulab-py:3.10-0.15.1` image and ssh into the session - it will fail because the `powerline-shell` is not found.

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -77,10 +77,10 @@ RUN apt-get update --fix-missing && \
 
 # inject the renku-jupyter stack
 COPY --from=renku_base /opt/conda /opt/conda
-COPY --from=renku_base --chown=rstudio:rstudio  /usr/local/bin/ /usr/local/bin/
-COPY --from=renku_base --chown=rstudio:rstudio \
-    /home/jovyan/ /home/rstudio/
+COPY --from=renku_base --chown=rstudio:rstudio /usr/local/bin/ /usr/local/bin/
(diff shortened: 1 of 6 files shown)


docker/r/Dockerfile
```diff
@@ -98,18 +98,18 @@ USER ${NB_USER}
 
 # set up conda in the NB_USER environment
 RUN echo ". ${CONDA_PATH}/etc/profile.d/conda.sh" >> ~/.bashrc && \
-    echo "conda activate base" >> ~/.bashrc
+    echo "conda activate base" >> ~/.bashrc && \
+    cat /renku/bashrc >> $HOME/.bashrc && \
+    echo "source ~/.bashrc" >> ~/.bash_profile
```
olevski, line 103: I think we need to do this in the main `py` image as well so that it gets applied to all other images?

Or is there a specific reason that this is only done in the R image but not anywhere else?

olevski (changes requested):
One minor question and potential change.

docker/r/Dockerfile
olevski: I am referring to `echo "source ~/.bashrc" >> ~/.bash_profile`

rokroskar: tbh I wasn't sure why the bash_profile was used in the R image - I presume it has something to do with the specifics of how R environments are set up. I just copied that line from another place in the Dockerfile, I didn't add it in this PR.

olevski (approved):

PR #329, opened 2023-04-03
chore: do not add provenance metadata to images

rokroskar:
Encoding provenance metadata in images seems to cause problems for many common docker libraries. See [here](https://community.fly.io/t/images-built-with-the-latest-docker-buildx-and-or-github-actions-work-again/10816) for an explanation.


rokroskar:
Tested this by pinning the image `renku/renkulab-py:3.10-cbcdd2e` in a project and the notebook service was able to find it.

olevski (approved):
Awesome! That works as expected Rok.

PR #345, opened 2023-05-19
chore: bump julia version

rokroskar:
Includes build for latest Julia release and parametrizes the build so other versions can be added later.


ciyer (approved):
I tested out the 1.9.0 image, and it works well. 🎉

PR #358, opened 2023-06-14
fix: use go-based powerline-shell

### LINKED ISSUE #357: Opening a shell prompt in a python session hangs
Details:
- image: renku/renkulab-py:3.9-0.17.0
- deployment: limited

Opening a terminal does not give a prompt immediately. It just hangs for a long time. Pressing ctrl-c during this time gives this error:

```
File "/opt/conda/bin/powerline-shell", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.9/site-packages/powerline_shell/__init__.py", line 240, in main
    segment.add_to_powerline()
  File "/opt/conda/lib/python3.9/site-packages/powerline_shell/segments/git.py", line 71, in add_to_powerline
    self.join()
  File "/opt/conda/lib/python3.9/threading.py", line 1060, in join
    self._wait_for_tstate_lock()
  File "/opt/conda/lib/python3.9/threading.py", line 1080, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
```

Therefore it is most likely powerline-shell causing this.

olevski:
Switch powerline shell to go.


olevski:
I tested this on the project where previously the terminal prompt was hanging. It does not behave this way with the new powerline shell.

olevski:
@rokroskar I ran the problematic project again (it is from Firat and it is not public). I cannot get the shell to hang at all. Even when I start the session immediately as soon as it is ready, then open one shell run `git lfs pull` and open other shells while the LFS files are downloaded. Also it seems that as soon as I ran `git lfs pull` I could also run `git status` successfully. Or perhaps the new sessions I launched got lucky and avoided whatever problem was causing `git status` to take forever to run.

As for needing the powerline fonts here is the relevant part from the repo/instructions:
https://github.com/justjanne/powerline-go#installation

Namely it says:

> If you want to use the "patched" mode (which is the default, and provides improved UI), you'll need to install a powerline font, either as fallback, or by patching the font you use for your terminal: see [powerline-fonts](https://github.com/Lokaltog/powerline-fonts). Alternatively you can use "compatible" or "flat" mode.

Out old config used `compatible` mode so I used compatible mode here as well.

rokroskar:
This does not fix #357 - I still get it to hang because the git prompt needs to call `git status` and git is rebuilding the index for some reason. But we should still switch to the powerline-go implementation, I'm all for that.

rokroskar:
For me, starting a local session hangs the terminal with this project: https://renkulab.io/projects/rok.roskar/adore 

I do `renku session start`, then first I got a message about "dubious repository ownership" and after fixing that I do `docker exec -ti <container-name> bash` and the shell hangs. Going into the container with `sh` I can run `top` and see

```
  891 jovyan    20   0  721448  13516   7340 S   5.3   0.1   1:20.70 git-lfs
 1057 jovyan    20   0  721704  13924   7596 S   5.3   0.1   1:10.31 git-lfs
 1134 jovyan    20   0  721448  14928   8440 S   5.0   0.1   1:08.39 git-lfs
  889 jovyan    20   0    7660   4148   3760 S   0.7   0.0   0:10.88 git
 1055 jovyan    20   0    7660   4148   3760 S   0.7   0.0   0:10.11 git
```

rokroskar (approved):

olevski:
@rokroskar funny enough using this definitely prevents the shell hanging in Firat's project in a session in the web ui. Because I can consistently get the terminal to hang when the old powerline shell is used and it never hangs with the new one.

PR #394, opened 2023-10-17
chore: R & RStudio to latest rocker with CRAN snapshot

RichardJActon:
The RStudio, R & Quarto versions in the default R 4.2.0 image are getting a little old now, especially the quarto version which is moving quite fast.

I tested a project based on an image built with these changes using the default R project template and it seems to work perfectly fine.

rocker/verse:devel gives R version 4.4.0 which is still the development version rather than the latest stable release so I switched it to 4.3.0 for verse and 2023.06.0-421 based on the latest versions from rocker https://github.com/rocker-org/rocker-versioned2/wiki/Versions for which there is a CRAN snapshot. That way other R package installs should have fixed versions based on that CRAN snapshot by default.

### DIFF
--- a/docker/r/Dockerfile
+++ b/docker/r/Dockerfile
@@ -1,6 +1,6 @@
 # define build arguments
 ARG RENKU_BASE=renku/renkulab-py:latest
-ARG BASE_IMAGE=rocker/verse:devel
+ARG BASE_IMAGE=rocker/verse:4.3.0
 
 # define base images
 FROM $RENKU_BASE as renku_base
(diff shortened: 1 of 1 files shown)


rokroskar:
Hi @RichardJActon sorry I missed this PR - I think we should combine it with #396? If it makes sense to you I'll take your commit over to the other PR so the CI will run.

docker/r/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # define build arguments
 ARG RENKU_BASE=renku/renkulab-py:latest
-ARG BASE_IMAGE=rocker/verse:devel
+ARG BASE_IMAGE=rocker/verse:4.3.0
```
rokroskar, line 3: this probably isn't needed since it's overridden in the CI anyway

rokroskar:
Hi @RichardJActon I've taken your commit over to #396 - will close this now and continue over there. Thanks again for making the updates!

PR #419, opened 2023-12-02
chore: add new bioc version

rokroskar:
adds latest bioconductor version

### DIFF
--- a/.github/workflows/build-and-push-to-docker.yml
+++ b/.github/workflows/build-and-push-to-docker.yml
@@ -549,6 +549,9 @@ jobs:
           - RVERSION: RELEASE_3_17
             BASE_IMAGE_NAME: bioconductor/bioconductor_docker
             EXTENSION: bioc
+          - RVERSION: RELEASE_3_18
+            BASE_IMAGE_NAME: bioconductor/bioconductor_docker
+            EXTENSION: bioc
           - RVERSION: devel
             BASE_IMAGE_NAME: bioconductor/bioconductor_docker
             EXTENSION: bioc
--- a/.github/workflows/build-and-push-to-docker.yml
+++ b/.github/workflows/build-and-push-to-docker.yml
@@ -552,9 +552,10 @@ jobs:
           - RVERSION: RELEASE_3_18
             BASE_IMAGE_NAME: bioconductor/bioconductor_docker
             EXTENSION: bioc
-          - RVERSION: devel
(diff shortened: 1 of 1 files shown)


rokroskar:
@RichardJActon any chance you might know why the R builds are suddenly failing? It's complaining that [`rlang` is not available](https://github.com/SwissDataScienceCenter/renkulab-docker/actions/runs/7268402444/job/19804391009?pr=419#step:8:2747)... any help much appreciated!

olevski:
Btw this is only a problem in `bioconductor/bioconductor_docker:devel`. And I think that we should simply stop publishing and using this version.

It wasn't fully clear to me but it uses an unstable version of R. This is what `R --version` says in `bioconductor/bioconductor_docker:devel`:

```
R Under development (unstable) (2023-12-13 r85679) -- "Unsuffered Consequences"
Copyright (C) 2023 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu
```

So if we keep using this we will probably be having little surprises like this once in a while. I manually reran all the official releases and they worked.

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -549,6 +549,9 @@ jobs:
           - RVERSION: RELEASE_3_17
             BASE_IMAGE_NAME: bioconductor/bioconductor_docker
             EXTENSION: bioc
+          - RVERSION: RELEASE_3_18
+            BASE_IMAGE_NAME: bioconductor/bioconductor_docker
+            EXTENSION: bioc
           - RVERSION: devel
             BASE_IMAGE_NAME: bioconductor/bioconductor_docker
             EXTENSION: bioc
```
olevski, line 557:
```suggestion
          # NOTE: The devel version of bioconductor uses unstable R so it often breaks or fails to build
          # - RVERSION: devel
          #    BASE_IMAGE_NAME: bioconductor/bioconductor_docker
          #    EXTENSION: bioc
```

olevski:
@rokroskar let me know if you agree to remove the `bioconductor/bioconductor_docker:devel` image. It would be nice to publish a new version of the templates. The new builds will include (automatically) patches for the terrapin SSH vulnerability.

rokroskar:
@olevski makes sense to me - I didn't realize that they use an unstable version of R.... 🙈

olevski (approved):

PR #465, opened 2024-08-20
chore: do not install server-proxy a second time

rokroskar:
jupyter server-proxy is already installed in the base renku image.


rokroskar:
superseded by #466

PR #476, opened 2024-11-13
feat: update images for new amalthea sessions

olevski:
Continuation of #473. I just did not want to push directly to your branch @leafty.


olevski:
I launched sessions with these new images here:
https://ci-renku-3712.dev.renku.ch/v2/projects/tasko.olevski/test-public-all-sessions

Everything works as expected. The only thing that users have to set in addition to the image is the `default_url`. I.e. if it is jupyter lab then they should set this to `/lab` or if it is an rstudio image then they should set this to `/rstudio`.

leafty:
Above issue is unrelated: https://github.com/SwissDataScienceCenter/amalthea/pull/768.

leafty (approved):
The new images seem to work well 🎉

PR #485, opened 2025-02-05
exp: run-image for python builds

leafty:


docker/vscodium-buildpacks/entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+HOME="${HOME:-/home/ubuntu}"
+RENKU_MOUNT_DIR="${RENKU_MOUNT_DIR:-${HOME}/work}"
+
+mkdir -p "${RENKU_MOUNT_DIR}/.vscode/extensions"
+
+# Install python extensions
+DEFAULT_VSCODIUM_EXTENSIONS="\
+  ms-python.python \
+  ms-toolsai.jupyter \
+"
```
rokroskar, line 14: This should be done in the docker image build, there's no reason to do it on launch

rokroskar (changes requested):
Thanks @leafty! Just a couple minor comments

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -928,3 +929,48 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3.2.0
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.DOCKER_PREFIX }}-buildpack-vscode-python
```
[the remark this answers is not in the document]
leafty: Yes, it's not a buildpacks, but it has very limited use outside of being used as a run image from buildpack builds.

docker/vscodium-buildpacks/entrypoint.sh
leafty: Why not?

By doing this at launch:
1. Any user can customize the extensions they want to have, just by tweaking the env vars.
2. You don't need to re-build to use more extensions.

The install time is very fast, so it doesn't seem like it would annoy users. This is similar to how devcontainers work.

rokroskar: Yes but those two will always be installed. So just put them in the image. And allow the user to specify additional ones if necessary.

.github/workflows/build-and-push-to-docker.yml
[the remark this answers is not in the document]
rokroskar: agreed, but calling it a "buildpack" carries a specific meaning, which is not accurate in this case so we should rename it.

docker/vscodium-buildpacks/entrypoint.sh
leafty: No, they will not alway be installed. If you specify `VSCODIUM_EXTENSIONS=" "` then no extension is installed, if you specify `VSCODIUM_EXTENSIONS="golang.go"` then you only have the go extension. The only issue at the moment is that custom env vars are not supported for Renku 2.0 sessions, so maybe this can be improved later to support command line arguments too.

rokroskar (approved):

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -928,3 +929,48 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3.2.0
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.DOCKER_PREFIX }}-vscodium-python-runimage:
```
leafty, line 959:
```suggestion
          images: ${{ env.DOCKER_PREFIX }}-vscodium-python-runimage
```

.github/workflows/build-and-push-to-docker.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           - docker/vscode/conda.Dockerfile
           - docker/vscode/poetry.Dockerfile
           - docker/vscode/R.Dockerfile
+          - docker/vscodium-buildpacks/base.Dockerfile
```
leafty, line 49:
```suggestion
          - docker/vscodium-runimage/python.Dockerfile
```

rokroskar (approved):

PR #490, opened 2025-02-24
fix: pin ubuntu + qemu for multi-arch builds

leafty:
See: https://github.com/actions/runner-images/issues/11471#issuecomment-2663514074


rokroskar (commented):
Can we just add a quick note to revert this change when it's fixed by GH? not sure how to track it, but ideally we wouldn't keep this pinned forever and forget about it.

leafty:
https://github.com/SwissDataScienceCenter/renkulab-docker/issues/491

rokroskar (approved):

